PR #19, opened 2024-05-22
Use SimpleHttpClient with proxy enabled, release v0.2.0

### LINKED ISSUE #18: Configured HTTP proxy not used when fetching CA roots
The HTTP client used for fetching CA roots does not respect the HTTP proxy configuration if it is present.

emgrav:
Closes #18

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -275,12 +275,12 @@ class InviteChecker:
         return resp.decode()
 
     async def _raw_gematik_root_ca_fetch(self) -> dict:
-        return await self.api.http_client.get_json(
+        return await self.api._hs.get_proxied_http_client().get_json(
             f"{self.config.gematik_ca_baseurl}/ECC/ROOT-CA/roots.json"
         )
 
     async def _raw_gematik_intermediate_cert_fetch(self, cn: str) -> bytes:
-        return await self.api.http_client.get_raw(
+        return await self.api._hs.get_proxied_http_client().get_raw(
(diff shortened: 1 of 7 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -275,12 +275,12 @@ async def _raw_federation_list_fetch(self) -> str:
         return resp.decode()
 
     async def _raw_gematik_root_ca_fetch(self) -> dict:
-        return await self.api.http_client.get_json(
+        return await self.api._hs.get_proxied_http_client().get_json(
```
tlater-famedly, line 278: We should at least leave a comment explaining why we're accessing a private member here; maybe we can get this to become a public interface upstream longer-term?

tests/rest.py
```diff
@@ -74,8 +74,7 @@ def create_room_as(
         expect_code: Literal[200] = ...,
         extra_content: dict | None = ...,
         custom_headers: Iterable[tuple[AnyStr, AnyStr]] | None = ...,
-    ) -> str:
```
tlater-famedly, line 77: Nit: noisy diff

tlater-famedly (approved):

sirewix (dismissed):

sirewix (dismissed):

sirewix (approved):

synapse_invite_checker/invite_checker.py
tlater-famedly: As mentioned by emily, we apparently access this all over the place already, so that ship has sailed.

Would still be nice to see if we can get the things we need to become public long-term, but that's beyond the scope of this PR. Let's move this to a separate issue?

PR #21, opened 2024-10-29
ci: Enable code coverage

emgrav:


lcov.info
```diff
@@ -0,0 +1,441 @@
+SF:synapse_invite_checker/__init__.py
```
mzaniolo, line 1: Do we really need this file or is this a mistake? This file was not on the other PRs so I'm not sure about it

mzaniolo (dismissed):

lcov.info
emgrav: no, this should be ignored, i just missed it. thank you

mzaniolo (approved):

PR #31, opened 2024-12-10
test: Verify newer Federation List schema is supported and test parsing of its data

### LINKED ISSUE #23: Use new federationlist version
### Description
We need to make sure to use the latest version of the list: https://github.com/gematik/api-vzd/blob/main/src/openapi/I_VZD_TIM_Provider_Services.yaml

### What has changed?
We need to update from version 1.3.1 to 1.3.7. The most important change is, that the domain scheme has changed. We need this new schema to make the /v1/server/findByIk endpoint work.

jason-famedly:
This will update the raw test data sample used, which required adjusting the domains that were tested against.

Three tests were added, to verify that data is parsed in the way expected. I found multiple servers that were not following the schema for the `timAnbieter` attribute, this is now tested for as well.

I think this:
* Fixes #23


synapse_invite_checker/invite_checker.py
```diff
@@ -352,12 +361,12 @@ async def user_may_invite(
         # Step 1, check federation allow list
         inviter_id = UserID.from_string(inviter)
         inviter_domain = inviter_id.domain
-        fedlist = await self.fetch_federation_allow_list()
-        if inviter_domain not in fedlist:
+        domain_set, _ = await self.fetch_federation_allow_list()
+        if inviter_domain not in domain_set:
             self.fetch_federation_allow_list.cache_clear()
-            fedlist = await self.fetch_federation_allow_list()
+            domain_set = await self.fetch_federation_allow_list()
```
nico-famedly, line 367: this looks wrong, if you now use a tuple

synapse_invite_checker/invite_checker.py
```diff
@@ -319,8 +327,9 @@ async def fetch_federation_allow_list(self) -> set[str]:
             msg = "Empty federation list"
             raise Exception(msg)
 
-        fedlist = FederationList.model_validate_json(jws_verify.payload)
-        return {fed.domain for fed in fedlist.domainList}
+        # Make sure to validate incoming, potentially incomplete or corrupt data
```
nico-famedly, line 330:
```suggestion
        # Validate incoming, potentially incomplete or corrupt data
```

nico-famedly: I think the other words are just filler here

tests/test_simple.py
```diff
@@ -865,3 +873,94 @@ def test_del_contact(self) -> None:
...
+        Search for a specific domain in the federation list
+        """
+        _, raw_fed_list = await self.invchecker.fetch_federation_allow_list()
+        assert len(raw_fed_list.domainList) > 0
+
+        for domain_entry in raw_fed_list.domainList:
+            if domain_entry.domain == domain:
+                return domain_entry
+
+        return None
```
nico-famedly, line 915: You can probably just assert here instead of returning None

nico-famedly:
Apart from the tuple confusion, this looks good enough for now, imo.

You may want to use a dataclass or some alternative to have named fields for the federation list return value, but that's something we can also tackle later.

Also it might make sense to keep the federation list tests together. You introduced a new testcase for it, but the old "is domain in list" assertion is still present in a different test. Probably makes sense to group them logically.

tests/test_simple.py
jason-famedly: I'm not certain what to `assert` here, and there may be a case for a test of what to do when a domain isn't present in the future. What do you recommend?

synapse_invite_checker/invite_checker.py
jason-famedly: good catch, I'll fix that up

synapse_invite_checker/invite_checker.py
jason-famedly: Yes this is not a school essay. Unnecessary words are unnecessary. They can go

tests/test_simple.py
nico-famedly: Well, we are always expecting to find those items in the tests which call it. So just assert for now and change it, if we ever have tests which expect None. Gets rid of some None checks at the caller side and makes the return type easier to read as well.

nico-famedly: So just `assert False, f"Not found in federation list {domain}"`

nico-famedly:
You would move the whole `test_federation_list` into the other test class for example and replaced the local invchecker, with the self.invchecker.

synapse_invite_checker/invite_checker.py
jason-famedly: Fixed in e64ae8f

synapse_invite_checker/invite_checker.py
jason-famedly: fixed in e64ae8f

tests/test_simple.py
jason-famedly: That sounds better than what I had in my head
fixed in e64ae8f

emgrav (commented):
One small note: We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) in the other development teams. It's up to the matrix team to decide for yourselves whether you also want to do this, but at least personally I am a sucker for consistency :) It has quite a few advantages, like allowing you to easily generate good changelogs and communicating intent more clearly at a glance.

nico-famedly (approved):
Looks good to me, thanks!

nico-famedly (approved):

PR #32, opened 2024-12-12
chore: Add matrix team to CODEOWNERS

nico-famedly:
The matrix team is probably now the primary party responsible for synapse and its modules. As such add us to the CODEOWNERS, so that we can review stuff. Not removing workflows for now, since the matrix team is quite small and it is very helpful, if sometimes externals can do reviews, when some of us are unavailable.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @famedly/workflows
+* @famedly/workflows @famedly/matrix-team


nico-famedly:
(For context, the Matrix team is 2 people rn, so any holiday would mean no reviewers for now, but feel free to scold me, if I am putting too much work onto workflows! :))

emgrav:
Sounds good. We will do what we can to review when necessary. Within the team we usually manage the review process out-of-band, so if there is a PR where you're dependent on a review from us, please post it in our matrix room so we don't miss it.

PR #34, opened 2024-12-16
ci: properly write coverage files in CI

nico-famedly:
By default we were only running the coverage command from the default environment, which doesn't write coverage files. The "cov" script in the "ci" environment does.

Also extend that script to write coverage.xml and to the terminal in addition to the lcov file, even though currently we only upload the lcov.info.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -2,10 +2,10 @@ name: Python
 
 on:
   push:
-    branches: [ "main" ]
-    tags: [ "*" ]
+    branches: ["main"]
+    tags: ["*"]
   pull_request:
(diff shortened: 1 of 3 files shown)


jason-famedly (approved):
I'm not completely certain what the `coverage.xml` file is supposed to accomplish yet, but generating it doesn't seem to harm anything. Don't forget to upload it if you wish to use it

Otherwise, LGTM :+1:

nico-famedly:
The coverage.xml is mostly just there, so that I don't have to figure out how to generate it in the future. Now all the files are there and we can decide, what we want to upload in the ci config instead of messing with hatch.

PR #36, opened 2024-12-18
refactor: Split unit tests into separate files

jason-famedly:
Break major test groups into separate files, with the goal of removing `test_simple.py`


tests/test_tim_federation.py
nico-famedly, line 1: This is more like the tim_federation_list, right?

nico-famedly (approved):
Seems a bit overly aggressive in splitting things up, but in general good, so I think we can merge it as is :)

tests/test_tim_federation.py
jason-famedly: For the moment, but it's generally names as there are more things coming and I don't want to have to re-do it

Issue #26, opened 2024-11-26
Implement isInsurance mode to block local invites and invites to other isInsurance server


nikzen:
### Description
With the new TI-Messenger Spec, there are two different kind of backends - TIM-Pro for Professionals and TIM-ePA for Patients. We need to implement a configuration option which allows this plugin to be used inside the TIM-Pro or TIM-ePA Backend.

### Solution
Implement config option tim-type to switch between pro or epa.

### ePA
- If enabled, the plugin needs to block all local invites.
- If enabled, the plugin needs to block all remote invites if the other server is type insurance (This information is stored on the federationList)

### Pro
- If enabled, the plugin needs to enable this endpoint GET /v1/server/isInsurance
- If enabled, the plugin needs to enable this endpoint GET /v1/server/findByIk

nico-famedly:
Strictly speaking we could also query the federation list to know, if we are an insurance and switch modes, but having a config option is probably easier :)

jason-famedly:
I have questions:
1. Is there any expectation for more 'modes' coming in the future? If not, a boolean may be appropriate instead of using `pro` and `epa`
2. Which one should be default? Allowing time for migration may be important
3. Does 'pro' mode include the same restrictions from 'epa' mode? In other words, when in 'pro' mode local invites are allowed and remote invites are allowed(but will be blocked by another server that is in 'epa' mode. Issue famedly/product-management#2385 may be involved in such case)?

nikzen:
> I have questions:
> 
> 1. Is there any expectation for more 'modes' coming in the future? If not, a boolean may be appropriate instead of using `pro` and `epa`
> 2. Which one should be default? Allowing time for migration may be important
> 3. Does 'pro' mode include the same restrictions from 'epa' mode? In other words, when in 'pro' mode local invites are allowed and remote invites are allowed(but will be blocked by another server that is in 'epa' mode. Issue [Bug: Trying to start a chat on a tim-server unauthorized opens an empty chat product-management#2385](https://github.com/famedly/product-management/issues/2385) may be involved in such case)?

1) There might be a tim-connect flag as well in future.
2) I think pro should be the default.
3) Local invites: Possible with pro, inpossible with epa. Remote invites: General possible within the gematik federation with pro & epa. If the invite is possible in the end is decided by the remote server, based on the user access list.

jason-famedly:
I found this:
[...]

PR #37, opened 2024-12-23
feat: Enable new

jason-famedly:
Part of #26 

Tasks:
* [ ] Parse `tim-type` as a new setting option. Update README
* [ ] Default `tim-type` to 'Pro' mode
* [ ] 'ePA' mode blocks all local invites
* [ ] Start saving comprehensive results from the parsed Federation List
* [ ] Implement serving new endpoint `/isInsurance` based on Federation Data
* [ ] Implement serving new endpoint `/findByIk` based on Federation Data
* Test new configuration setting:
   * [ ] Parsing is case-insensitive, defaults to 'Pro' mode, and(if incorrectly set) raises an Error
* 'ePA' mode 
   * [ ] Test all local invites are blocked
   * [ ] Test outgoing remote invites are blocked if remote server is an 'Insurance' type(What do we do for incoming remote invites from same? I guess they are allowed so Insurance can contact a patient)
   * [ ] Test new `/isInsurance` and `/findByIk` return expected error code 401
* 'Pro' mode
   * [ ] Test all local invites are not blocked
   * [ ] Test outgoing remote invites are allowed
   * [ ] Test new `/isInsurance` and `/findByIk` return correct data


README.md
```diff
@@ -37,6 +37,7 @@ modules:
         federation_localization_url: "https://localhost:8080/localization", # Full url where to fetch the federation localization from, required. Should be the same host as federation list.
         federation_list_client_cert: "tests/certs/client.pem", # path to a pem encoded client certificate for mtls, required if federation list url is https
         gematik_ca_baseurl: "https://download-ref.tsl.ti-dienste.de/", # the baseurl to the ca to use for the federation list, required
+        tim-type: "epa" or "pro", # Patient or Professional mode, optional. Defaults to "pro"
```
nico-famedly, line 40:
```suggestion
        tim-type: "epa" or "pro", # Patient/Insurance or Professional mode, optional. Defaults to "pro"
```

I think the epa mode is really to run the insurance services to give patients access to TIM, so we probably should name them as well?

jason-famedly: We can, but only patients can be insured. I felt it was implied terminology. If it's a doctor being seen, he's a 'patient' in that context and has to go thru his insurance to connect! (Which, to be fair, I see as not happening when they can take the 'already established route and just message another doctor)

synapse_invite_checker/invite_checker.py
```diff
@@ -296,13 +321,17 @@ def _load_cert_b64(self, cert: str) -> X509:
         return load_certificate(FILETYPE_ASN1, base64.b64decode(cert))
 
     @cached(cache=TTLCache(maxsize=1, ttl=60 * 60))
-    async def fetch_federation_allow_list(self) -> tuple[set[str], FederationList]:
+    async def fetch_federation_allow_list(
+        self,
+    ) -> tuple[set[str], set[str], FederationList]:
```
nico-famedly, line 326: This should really be a class at this point, this is getting hard to parse :)

synapse_invite_checker/invite_checker.py
```diff
@@ -179,14 +183,17 @@ def __init__(self, config: InviteCheckerConfig, api: ModuleApi):
 
         # The TiMessengerInformation API resource
         MessengerInfoResource(self.config).register(self.resource)
+        MessengerIsInsuranceResource(
+            self.config, self.is_domain_insurance_with_retry
```
nico-famedly, line 187: Imo it is easier to read, if you call that directly on the InviteChecker instead of passing in a callback?

README.md
nico-famedly: This should also mention, that this will be required in the future, so people really should set it?

synapse_invite_checker/rest/messenger_info.py
```diff
@@ -42,11 +43,36 @@ def __init__(self, config: InviteCheckerConfig):
...
+        self.PATTERNS = tim_info_patterns("/v1/server/isInsurance$")
+
+    async def on_GET(self, request: SynapseRequest) -> tuple[int, JsonDict]:
+        if self.config.tim_type == TimType.EPA:
+            return 401, {
+                "errorCode": "thing",
+                "errorMessage": "Unauthorized to access for ePA backend",
+            }
+
+        server_name = parse_string(request, "serverName")
```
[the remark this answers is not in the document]
jason-famedly: Yes, good catch

tests/test_local_invites.py
```diff
@@ -35,7 +42,7 @@ def test_invite_to_dm(self) -> None:
         room_id = self.helper.create_room_as(
             self.user_a, is_public=False, tok=self.access_token
         )
-        assert room_id, "Room created"
+        assert room_id, "Room not created"
```
nico-famedly, line 45: Wouldn't now the error be: "Assertion Room not created failed"? That seems backwards to me.

synapse_invite_checker/invite_checker.py
jason-famedly: Maybe, but this spells out exactly what function is associated with that rest endpoint without looking at the resource construction. I didn't want to just pass the whole InviteChecker in when all that is needed from it is the config and the one function to call

synapse_invite_checker/invite_checker.py
jason-famedly: I'll consider it in a following round. There are a couple different angles and variables to take into account and it needs discussion

synapse_invite_checker/rest/messenger_info.py
```diff
@@ -42,11 +43,36 @@ def __init__(self, config: InviteCheckerConfig):
...
+        self.is_insurance_cb = is_insurance_cb
+
+        self.PATTERNS = tim_info_patterns("/v1/server/isInsurance$")
+
+    async def on_GET(self, request: SynapseRequest) -> tuple[int, JsonDict]:
+        if self.config.tim_type == TimType.EPA:
+            return 401, {
+                "errorCode": "thing",
+                "errorMessage": "Unauthorized to access for ePA backend",
+            }
```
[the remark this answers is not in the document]
jason-famedly: I think I got held up there by going to look for the verbiage on the `errorMessage`, not finding it and instead discovering several other rabbit holes of semi-related things. Good catch

tests/test_local_invites.py
```diff
@@ -121,3 +128,122 @@ def test_invite_to_group_without_dm_event(self) -> None:
...
+    These tests operate under the assumption that invites are processed after room
+    creation has occurred. Invites during room creation will be tested separately
+    """
+
+    def prepare(self, reactor: MemoryReactor, clock: Clock, homeserver: HomeServer):
+        super().prepare(reactor, clock, homeserver)
+        self.user_a = self.register_user("a", "password")
+        self.access_token = self.login("a", "password")
+        self.user_b = self.register_user("b", "password")
+        self.user_c = self.register_user("c", "password")
```
jason-famedly, line 144: That reminds me tho, need to fix these. These users can't exist on an insurance domain(I think, I might be mis-remembering)

synapse_invite_checker/invite_checker.py
```diff
@@ -362,21 +422,27 @@ async def user_may_invite(
...
+        if (
+            self.config.tim_type == TimType.EPA
+            and await self.is_domain_insurance_with_retry(inviter_domain)
+        ):
+            # Per AF_10233: Deny incoming remote invites if in ePA mode(which means the
+            # local user is an 'insured') and if the remote domain is type 'insurance'.
+            logger.warning(
+                "Discarding invite from remote insurance domain: %s", inviter_domain
+            )
+            return errors.Codes.FORBIDDEN
```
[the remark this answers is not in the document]
jason-famedly: You know what? You may be right. I was passing that point of view on the parent issue, but the spec *does* say to check both us and them.
https://gemspec.gematik.de/docs/gemSpec/gemSpec_TI-M_ePA/latest/#A_25705

I'll get this sorted out soon

jason-famedly:
squashing up to this point, so new work will be after that

synapse_invite_checker/invite_checker.py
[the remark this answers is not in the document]
nico-famedly: In the end the other side will check the federation list when validating the invite. As such if the admin sets something else, it will break randomly and hard to diagnose. That is why I would make a noisy warning on startup if the setting is different than what we would expect from the federation list.

Now if we should check epa or the list when doing the actual checks, I don't know. I would probably prioritize the list, but as long as the admin sets the setting correctly, it shouldn't matter? Which is why I think warning on startup makes sense?

synapse_invite_checker/invite_checker.py
```diff
@@ -337,14 +366,45 @@ async def fetch_federation_allow_list(self) -> tuple[set[str], FederationList]:
...
+    async def is_domain_insurance_with_retry(self, domain: str) -> bool:
+        _, insurance_domains, _ = await self.fetch_federation_allow_list()
+        if domain in insurance_domains:
+            return True
+
+        # The domain wasn't found but the list may have changed since the last look.
+        # Re-fetch the list and try again
+        self.fetch_federation_allow_list.cache_clear()
+        _, insurance_domains, _ = await self.fetch_federation_allow_list()
+        return domain in insurance_domains
```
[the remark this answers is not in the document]
jason-famedly: See what you think about [413381a](https://github.com/famedly/synapse-invite-checker/pull/37/commits/413381a2ee750c7deeae497d0ffc2d47c2704be3)

synapse_invite_checker/invite_checker.py
jason-famedly: I think I did that in [413381a](https://github.com/famedly/synapse-invite-checker/pull/37/commits/413381a2ee750c7deeae497d0ffc2d47c2704be3) but it may not have been what you meant

synapse_invite_checker/invite_checker.py
```diff
@@ -296,13 +343,17 @@ def _load_cert_b64(self, cert: str) -> X509:
...
         """
         Fetch the raw data for the federation list, verify it is authentic and parse
         the data into a usable format
 
-        Returns: a tuple of a set of the domains from the list and the complete
-            validated list(for testing)
+        Returns:
+            a 3-tuple of a set of the domains from the list,
+            a set of domains that are insurance,
+            and the complete validated list(for testing)
```
nico-famedly, line 356: This seems wrong? :)

synapse_invite_checker/types.py
```diff
@@ -61,3 +62,28 @@ class FederationList(BaseModel):
...
+    @computed_field
+    @cached_property
+    def _insurance_domains_on_list(self) -> list[str]:
+        return [
+            domain_data.domain
+            for domain_data in self.domainList
+            if domain_data.isInsurance
+        ]
+
+    def allowed(self, domain: str) -> bool:
```
nico-famedly, line 80: I know I used that term out of laziness, but probably should be "federation_allowed" or have some documentation?

tests/test_local_invites.py
```diff
@@ -12,21 +12,32 @@
...
+    """
+
     def prepare(self, reactor: MemoryReactor, clock: Clock, homeserver: HomeServer):
         super().prepare(reactor, clock, homeserver)
         self.user_a = self.register_user("a", "password")
         self.access_token = self.login("a", "password")
         self.user_b = self.register_user("b", "password")
         self.user_c = self.register_user("c", "password")
 
+        # @d:test is none of these things and should be just a 'User'
```
nico-famedly, line 37: None of what? Not a practitioner?

synapse_invite_checker/invite_checker.py
jason-famedly: Good catch, fixed

synapse_invite_checker/types.py
```diff
@@ -61,3 +62,28 @@ class FederationList(BaseModel):
     )
 
     domainList: list[FederationDomain]  # noqa: N815
+
+    @computed_field
+    @cached_property
+    def _domains_on_list(self) -> list[str]:
```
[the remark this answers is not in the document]
jason-famedly: I think they can be, the sequence itself doesn't have to have a concrete order. Fixed

synapse_invite_checker/types.py
jason-famedly: I added some docstrings

tests/test_remote_invites.py
```diff
@@ -162,24 +236,350 @@ async def test_invite_from_remote_to_local_org(self) -> None:
...
+        if ret == NOT_SPAM:
+            return NOT_SPAM
+        return ret[0]  # return first code instead of all of them to make assert easier
+
+    async def test_invite_from_remote_not_on_fed_list(self) -> None:
+        """Tests that an invite from a remote server not in the federation list gets denied"""
+
+        channel = self.make_request(
+            "PUT",
+            "/_synapse/client/com.famedly/tim/v1/contacts",
```
nico-famedly, line 344: I'm not sure the contacts API should be available in EPA mode?

tests/test_local_invites.py
jason-famedly: Yes. So like a nurse or a doctor's secretary. From what I can understand from the spec these people are considered Users. Like an office manager or something. I updated the comment to add what each user a-c is since that was a fair bit of missing context(In the other similar places too)

tests/test_remote_invites.py
```diff
@@ -162,24 +236,350 @@ async def test_invite_from_remote_to_local_org(self) -> None:
...
+        self.helper.auth_user_id = self.user_d
+
+    def default_config(self) -> dict[str, Any]:
+        conf = super().default_config()
+        assert "modules" in conf, "modules missing from config dict during construction"
+
+        # There should only be a single item in the 'modules' list, since this tests that module
+        assert len(conf["modules"]) == 1, "more than one module found in config"
+
+        conf["modules"][0].setdefault("config", {}).update({"tim-type": "epa"})
```
nico-famedly, line 324: does that work properly if our hs name is "test"?

tests/test_local_invites.py
```diff
@@ -121,3 +146,136 @@ def test_invite_to_group_without_dm_event(self) -> None:
...
+            targ=self.user_d,
+            tok=self.access_token,
+            expect_code=200,
+        )
+
+
+class LocalEpaModeInviteTest(ModuleApiTestCase):
+    """
+    These tests operate under the assumption that invites are processed after room
+    creation has occurred. Invites during room creation will be tested separately
```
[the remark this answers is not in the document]
jason-famedly: I don't actually believe they do, at least when it comes the follow up it didn't seem to. I spelled this out in the context of the issue that mentioned changing the SDK to only do invites after room creation(even tho that was for invites of additional users after the fact)

tests/test_remote_invites.py
jason-famedly: It's a gray area for me. I thought that if I was a patient and I needed to contact my doctor who was not a publicly listed practitioner, then I needed their contact details?

tests/test_remote_invites.py
jason-famedly: It does because we just warn now if there is a conflict in our two sources of "truth"

tests/test_local_invites.py
nico-famedly: You use such users in other tests, but not in these. So I think you just copy and pasted that user and don't need them here?

tests/test_local_invites.py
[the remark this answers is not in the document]
nico-famedly: I am pretty sure I tested this and the spam check endpoint does get invoked during room creation. The follow up is about something else, where multiple invites during room creation can't reliably return errors.

tests/test_remote_invites.py
nico-famedly: The contact API is about making yourself contactable. Patients don't need that and they can't use that to contact doctors either, but double check the spec I guess?

tests/test_remote_invites.py
nico-famedly: That's not really intended functionality though. I guess it works, but should have a comment then.

tests/test_local_invites.py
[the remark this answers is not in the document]
jason-famedly: And yet doctors are allowed to invite additional users to a room after the fact. So, I guess I don't understand what you are asking?

tests/test_local_invites.py
jason-famedly: Oh I have their access token. I can lose that, or I can write an additional iteration of tests to check that scenario. Which would you prefer?

tests/test_local_invites.py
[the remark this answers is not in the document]
jason-famedly: > I am pretty sure I tested this and the spam check endpoint does get invoked during room creation. The follow up is about something else, where multiple invites during room creation can't reliably return errors.

You may have tested this, but it is not currently tested here

tests/test_remote_invites.py
jason-famedly: I spent(far to much) time examine the spec and could find nothing that spells out that a patient is not allowed to have a contact permission feature. I think the contact permissions here are incomplete though, I'm seeing a "Block All" but allow exceptions and the inverse is missing. Let's kick that further down the road as it is creeping into a larger problem.

synapse_invite_checker/invite_checker.py
```diff
@@ -362,21 +434,26 @@ async def user_may_invite(
...
+        ):
+            # Per AF_10233: Deny incoming remote invites if in ePA mode(which means the
+            # local user is an 'insured') and if the remote domain is type 'insurance'.
+            logger.warning(
+                "Discarding invite from remote insurance domain: %s", inviter_domain
+            )
+            return errors.Codes.FORBIDDEN
 
         # Step 2, check invite settings
         contact = await self.store.get_contact(UserID.from_string(invitee), inviter)
```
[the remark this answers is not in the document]
jason-famedly: combine with other discussion around this area
https://github.com/famedly/synapse-invite-checker/pull/37#discussion_r1908582012

tests/test_remote_invites.py
jason-famedly: I added such a comment in both places I thought it was relevant

tests/test_local_invites.py
jason-famedly: I just removed the access token for that user. It wasn't actually being used. May add tests for that later

synapse_invite_checker/invite_checker.py
```diff
@@ -179,14 +185,22 @@ def __init__(self, config: InviteCheckerConfig, api: ModuleApi):
...
+        MessengerIsInsuranceResource(self.config, self.is_domain_insurance).register(
+            self.resource
+        )
 
         # Register everything at the root of our namespace/app, to avoid complicating
         # Synapse's regex http registration
         self.api.register_web_resource(BASE_API_PREFIX, self.resource)
+
+        # Install a callable that will be registered to run after the reactor has been
+        # completely setup and has started running.
```
nico-famedly, line 197: That part is obvious, if you want to document something, document why we are doing that, i.e.:

> We can only validate the EPA mode once the HS is running.

synapse_invite_checker/invite_checker.py
```diff
@@ -223,8 +237,41 @@ def parse_config(config):
...
+            else:
+                msg = "`tim-type` setting is not a recognized value. Please fix."
+                raise ConfigError(msg)
+
         return _config
 
+    def after_startup(self) -> None:
+        """
+        Attach to the reactor, and after the reactor is fully set up and running
+        check the federation list to see if if the current server isInsurance
```
nico-famedly, line 263:
```suggestion
        To be called when the rector is running. Validates that the epa setting matches the insurance setting in the federation list.
```

synapse_invite_checker/invite_checker.py
```diff
@@ -362,23 +434,33 @@ async def user_may_invite(
...
-                logger.warning("Discarding invite from domain: %s", inviter_domain)
-                return errors.Codes.FORBIDDEN
+        # Step 1a, check federation allow list
+        if not await self.is_domain_allowed(inviter_domain):
+            logger.warning("Discarding invite from domain: %s", inviter_domain)
+            return errors.Codes.FORBIDDEN
+
+        # Step 1b, check that both domains are not of type 'epa'
+        # TODO: stop trusting our local setting and instead check the federation list
+        #  data to see if *this* server is of type `isInsurance`
```
nico-famedly, line 450: Is this a todo? I thought we intentionally use the epa setting?

```suggestion
        # We warn during startup if the 'epa' setting and isInsurance are not equivalent, so we can use the setting to know, if we are an insurance.
```

nico-famedly: this seems to be repeated later, so you can probably merge those comments.

nico-famedly (dismissed):
Code seems fine, some of the comments still are a bit redundant to me, but whatever :)

synapse_invite_checker/invite_checker.py
jason-famedly: I'll remove it

synapse_invite_checker/invite_checker.py
jason-famedly: Sounds good, I'll use that

Issue #35, opened 2024-12-16
Need to implement method to verify that only a single additional invite can be included when calling `createRoom`


jason-famedly:
Per the third paragraph here: https://github.com/gematik/api-ti-messenger/blob/main/docs/Fachdienst/MessengerService.adoc#21-client-server-pr%C3%BCfungen

> The following checks must be carried out by the messenger proxy :
> - The messenger proxy must ensure that the attribute is filled `invite` with a maximum of one MXID when a room is created.
> - For each request, the messenger proxy `Invite-Event` must check whether the matrix domains contained in the request belong to the TI federation.
> - For incoming communication to the media endpoints, the messenger proxy must check the path component {serverName} for federation membership

(Translation by Google, formatting is mine)

The first of these should be implemented in this code base. Out-of-band discussion with @nico-famedly has led to the conclusion that a third party rules callback for [`on_create_room()`](https://element-hq.github.io/synapse/latest/modules/third_party_rules_callbacks.html#on_create_room) would be an appropriate method to handle this.

PR #39, opened 2025-01-08
feat: Implement local room creation limitations with included invites

jason-famedly:
This may be a breaking change, let me know. If it is, I can include a "disabler" config option to allow time to migrate.

According to [A_25322](https://gemspec.gematik.de/docs/gemSpec/gemSpec_TI-M_Basis/gemSpec_TI-M_Basis_V1.1.1/#A_25322):
> When calling the API to create a room, the Matrix home server MUST only allow the invitation of a maximum of one MXID.

This work implements that limitation before a room is even created. Additionally, [`A_25705`](https://gemspec.gematik.de/docs/gemSpec/gemSpec_TI-M_ePA/latest/#A_25705) mandates that:
> The messenger proxy of the TI-M FD ePA MUST check requests for invite events as part of the client-server checks. If the sender and the recipient are both actors in the role of insured person, the invitation MUST be rejected by the TI messenger proxy. An actor is to be identified as an insured person if the domain of his MXID has the value " true " in the field " isInsurance " within the federation list.

And has been included as well


synapse_invite_checker/invite_checker.py
```diff
@@ -514,3 +521,64 @@ async def user_may_invite(
...
+        # Interesting potential error here, they display an http error code of 400, but
+        # then say to use "M_FORBIDDEN". Pretty sure that is a typo
+        if len(invites_block) > 1:
+            raise SynapseError(
+                403,
+                "When creating a room, a maximum of one participant can be invited directly",
+                errors.Codes.FORBIDDEN,
+            )
+
+        for invitee_entry in invites_block:
```
nico-famedly, line 552: This is redundant, Synapse calls the invite spam checks already during room creation.

nico-famedly: If you want better error messages, there is still no reason to copy all of this code :p

jason-famedly: Sorted it here [1acb167](https://github.com/famedly/synapse-invite-checker/pull/39/commits/1acb167954a4c0a4d0ccbb038e92271c07a15bd5)

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
...
+    ) -> None:
+        """
+        Raise a SynapseError if creating a room with invites should be denied
+        """
+        # Unlike `user_may_invite()`, `on_create_room()` runs with the inviter being
+        # a local user and the invitee is remote
+        inviter = requester.user
+        assert self.api.is_mine(
+            inviter.to_string()
+        ), "creating a room for a user that wasn't from this server is strange"
```
jason-famedly, line 426: I can get rid of the comment and the assert here, as the assert isn't necessary(I think) and the comment is slightly wrong

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
...
     async def user_may_invite(
         self, inviter: str, invitee: str, room_id: str
     ) -> Literal["NOT_SPAM"] | errors.Codes:
-        # Check local invites first, no need to check federation invites for those
-        if self.api.is_mine(inviter):
+        try:
+            await self._check_invite(inviter, invitee, room_id)
+        except SynapseError as e:
+            if e.code == 403:
+                logger.debug(f"USER_MAY_INVITE ERROR: {e.msg}")
```
jason-famedly, line 456: I'm open to a better DEBUG message here

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
...
-                            "Preventing invite since %s already has a DM with %s",
-                            inviter,
-                            invitee,
-                        )
-                        return errors.Codes.FORBIDDEN
-
-            # local invites are always valid, if they are not to a dm.
-            # There are no checks done for outgoing invites apart from in the federation proxy,
-            # which checks all requests.
+            if room_id:
```
jason-famedly, line 475: In context, when running the logic during the room creation process there is not a room_id yet.

Also, as part of the combined logic above, this is no longer checked between a local and a remote user. I can duplicate it below in the appropriate section if it is still needed. Or(if I understand correctly) it can go if it is now redundant

jason-famedly: It might also be a possibility that we don't check the room_id at all anymore and only observer that if the two users share a DM at all(regardless of the room_id) that it is a situation where we forbid

synapse_invite_checker/invite_checker.py
nico-famedly: It's probably fine, the assert isn't called when optimized anyway, afaik. Just make the assertion "Only local users can create rooms" or so.

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
     async def is_domain_insurance(self, domain: str) -> bool:
         return await self._domain_list_check(lambda fl: fl.is_insurance(domain))
 
+    async def on_create_room(
+        self,
+        requester: Requester,
+        request_content: dict[str, Any],
+        is_request_admin: bool,
```
nico-famedly, line 416: you should probably do nothing if the inviter is an admin.

nico-famedly: (To match what the invite spam checks usually do)

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
...
+    ) -> None:
+        """
+        Raise a SynapseError if creating a room with invites should be denied
+        """
+        # Unlike `user_may_invite()`, `on_create_room()` runs with the inviter being
+        # a local user and the invitee is remote
```
nico-famedly, line 422: What? user_may_invite is executed for both remote invites and local invites. This comment makes no sense, unless you meant to say "for create_room the inviter is always the local user".

Also you may wanna explain why we are not using user_may_create_room.

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
...
+        inviter = requester.user
+        assert self.api.is_mine(
+            inviter.to_string()
+        ), "creating a room for a user that wasn't from this server is strange"
+
+        invites_block: list[str] = request_content.get("invite", [])
```
nico-famedly, line 428:
```suggestion
        invite_list: list[str] = request_content.get("invite", [])
```

No idea, why that would be a block.

synapse_invite_checker/invite_checker.py
```diff
@@ -404,44 +409,100 @@ async def is_domain_allowed(self, domain: str) -> bool:
...
+                errors.Codes.FORBIDDEN,
+            )
+
+        #  The spec says this is to be checked by the messenger proxy. I believe this
+        #  leaves a dangling room with only the original creator and no real evidence
+        #  that the other party will not be joining. Logically, blocking the room
+        #  creation in the situation where we already know the other party can't/won't
+        #  attend is cleaner
+        for invitee_entry in invites_block:
+            await self._check_invite(inviter.to_string(), invitee_entry)
```
nico-famedly, line 447: Why not do what synapse does?

```suggestion
        for invitee in invite_list:
            res = await self._check_invite(inviter.to_string(), invitee_entry)
            if res != NOT_SPAM:
              raise SynapseError...
```

What you are doing below of making the invite checks potentially raise an exception and then mapping it back is really confusing.

synapse_invite_checker/invite_checker.py
nico-famedly: Why even have a debug message there? Do you just want to log, that an invite was blocked?

synapse_invite_checker/invite_checker.py
nico-famedly: No, we should verify that if it is a known room, that we don't invite into the dm. This can't happen during room creation anyway. You should comment, that this is why we skip the check, I guess.

nico-famedly (changes requested):

tests/base.py
```diff
@@ -135,7 +136,7 @@ def return_localization(mxid: str) -> str:
         f"matrix:u/matrixuri2:{DOMAIN_IN_LIST}",
         f"matrix:user/gematikuri%3A{DOMAIN_IN_LIST}",
         f"matrix:user/gematikuri2:{DOMAIN_IN_LIST}",
-        "matrix:u/a%3Atest",
+        f"matrix:u/a%3A{SERVER_NAME}",
```
nico-famedly, line 139: This doesn't match our actual mxids now anymore, I think?

synapse_invite_checker/invite_checker.py
jason-famedly: This was the comment I was referencing at https://github.com/famedly/synapse-invite-checker/pull/39/files#r1915174243

According to the module api docs, [`user_may_create_room()`](https://element-hq.github.io/synapse/latest/modules/spam_checker_callbacks.html#user_may_create_room) can only accept the information of the `user_id` and has no other reference available to compare if another user is being invited

synapse_invite_checker/invite_checker.py
jason-famedly: The arg isn't used, but was required by the signature

synapse_invite_checker/invite_checker.py
jason-famedly: removed it

synapse_invite_checker/invite_checker.py
jason-famedly: no idea either, something I missed at the time. Fixed now

synapse_invite_checker/invite_checker.py
jason-famedly: removed all related SynapseErrors. But, for the record, we don't want to raise a SynapseError directly through `user_may_invite()`. It would technically work, but borks up the machinery in Synapse that is expecting to keep running other callbacks. Like I said, it would work, but it feels so dirty

synapse_invite_checker/invite_checker.py
jason-famedly: it's gone now

synapse_invite_checker/invite_checker.py
jason-famedly: It's gone. This doesn't run during room creation anymore

tests/base.py
jason-famedly: it doesn't(and it got renamed so I hope this doesn't get lost). The test now assumes the identity of a server domain from the federation list so that testing can be done as it would be in the wild. More 'real' if you want

jason-famedly:
Start over now please. Most of it is different now

jason-famedly:
Resolves #35

synapse_invite_checker/invite_checker.py
```diff
@@ -430,17 +462,26 @@ async def user_may_invite(
                         return errors.Codes.FORBIDDEN
 
             # local invites are always valid, if they are not to a dm.
```
nico-famedly, line 464:
```suggestion
            # local invites are always valid, if they are not to a dm (and not in EPA mode).
```

synapse_invite_checker/invite_checker.py
nico-famedly: Yeah, I didn't want you to raise from user_may_invite, what you did now looks good :)

jason-famedly: Alright. How about that?

nico-famedly (approved):

nico-famedly:
Looks good to me, thanks!

PR #40, opened 2025-01-22
Contact REST Api migration

jason-famedly:
Fixes:
* #25 

Setup a new handler for permissions in Synapse's account data. 
* [x] Stop using the database to store new contact entries
* [x] On first start up, migrate all contact entries to account_data and the new permissions handler.
  * <b>See data loss disclaimer below.</b>
  * <b>NOTE: This effectively makes that user's data instantly available to the client app as account data.</b>
* [x] The Contacts REST API can continue to be used only in "block all" mode, provided that adding a new user is the only change. <b>See data loss disclaimer below.</b>
* [x] As migration takes place for a user the database entries for that user will be deleted
* [x] new tests

Disclaimers:
* I do not know what to do about the data loss that will occur because of this. The hint of `displayName` is the more concerning of these. Does the client app have a way to find that piece of information even if they do not have access to the public profile on the remote server(excepting when there is already a shared room between the users, which feels like a dead end scenario). Loss of `start` and `end` times in the `InviteSettings` I'm more ok with.
* <del>Upstream concern: The account data being set requires an "event type" to be set. For ePA and HBA users, this has been described in the gematik spec. For Users that are neither of those, there is no "event type" established. I have generated one to use for now based on surround criteria</del> I mis read how this works, ignore
* <del>Upstream concern: I have witnessed circumstantial evidence saying there are times that an HBA may not have used their HBA access card(?) to access their account when logging in. If this is true, then I am not sure if there is a concern or not that they may not be on the localization data as `pract` which is a requirement for being a publicly visible HBA. That I know of, I have on other way to discover which type of User is being used to determine the "event type" described immediately above. Might there be an issue where the new permissions in account data won't become accessible if this switch changes?</del> Same reason as above

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -15,7 +15,6 @@
 import base64
 import functools
 import logging
-import time
 from contextlib import suppress
 from typing import Any, Callable, Dict, Literal, cast
 from urllib.parse import quote, urlparse
@@ -48,6 +47,9 @@ from twisted.web.iweb import IPolicyForHTTPS
 from zope.interface import implementer
 
 from synapse_invite_checker.config import InviteCheckerConfig
+from synapse_invite_checker.permissions import (
+    InviteCheckerPermissionsHandler,
+)
 from synapse_invite_checker.rest.contacts import (
     ContactManagementInfoResource,
     ContactResource,
@@ -182,10 +184,20 @@ class InviteChecker:
         ) as db_conn:
             self.store = InviteCheckerStore(api._store.db_pool, db_conn, api._hs)
 
+        self.permissions_handler = InviteCheckerPermissionsHandler(
+            self.api,
+            self.fetch_localization_for_mxid,
+            self.is_domain_insurance,
+        )
+
         # The Contact Management API resources
         ContactManagementInfoResource(self.config).register(self.resource)
-        ContactsResource(self.api, self.store).register(self.resource)
-        ContactResource(self.api, self.store).register(self.resource)
+        ContactsResource(self.api, self.store, self.permissions_handler).register(
+            self.resource
+        )
+        ContactResource(self.api, self.store, self.permissions_handler).register(
+            self.resource
+        )
 
         # The TiMessengerInformation API resource
         MessengerInfoResource(self.config).register(self.resource)
@@ -261,8 +273,9 @@ class InviteChecker:
 
     def after_startup(self) -> None:
         """
-        To be called when the rector is running. Validates that the epa setting matches
-        the insurance setting in the federation list.
+        To be called when the reactor is running. Validates that the epa setting matches
+        the insurance setting in the federation list and *might* perform forced contact
+        migration.
         """
         fetch_deferred = Deferred.fromCoroutine(self._fetch_federation_list())
         fed_list = cast(FederationList, fetch_deferred.result)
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -38,6 +38,7 @@ modules:
         federation_list_client_cert: "tests/certs/client.pem", # path to a pem encoded client certificate for mtls, required if federation list url is https
         gematik_ca_baseurl: "https://download-ref.tsl.ti-dienste.de/", # the baseurl to the ca to use for the federation list, required
         tim-type: "epa" or "pro", # Patient/Insurance or Professional mode, defaults to "pro" mode. Optional currently, but will be required in a later release
+        force-migration: boolean, # Default to false. Recommend not using for a couple weeks until on-demand migration has occurred
```
nico-famedly, line 41: Maybe describe also, what this is for :3

synapse_invite_checker/invite_checker.py
```diff
@@ -109,7 +111,9 @@ def __init__(self, config: InviteCheckerConfig):
                 content = file.read()
 
             client_cert = PrivateCertificate.loadPEM(content)
-        elif self.url.scheme != "http":
+        # pragma no cover -  We don't run coverage for this as the condition is caught
```
nico-famedly, line 114: Do we need the pragma though? I think it is fine to just have it explicitly not covered as other code might change some day and we need to recheck if this is a good or bad idea to have this line covered?

synapse_invite_checker/rest/contacts.py
```diff
@@ -60,18 +64,29 @@ async def on_GET(self, _: SynapseRequest) -> tuple[int, JsonDict]:
...
         self.api = api
+        self.permission_handler = permission_handler
         self.PATTERNS = contact_mgmt_patterns("/contacts$")
 
-    # @override
     async def on_GET(self, request: SynapseRequest) -> tuple[int, JsonDict]:
         requester = await self.api.get_user_by_req(request)
+        await self.permission_handler.do_migration(self.store, requester.user)
+
+        contacts = await self.permission_handler.get_contacts(
```
nico-famedly, line 83: I think this should return an error if the user flipped the allow_all bit, because then this API shouldn't be used anymore imo?

synapse_invite_checker/rest/contacts.py
```diff
@@ -94,17 +112,25 @@ async def on_PUT(self, request: SynapseRequest) -> tuple[int, JsonDict]:
...
         super().__init__()
         self.store = store
         self.api = api
+        self.permission_handler = permission_handler
         self.PATTERNS = contact_mgmt_patterns("/contacts/(?P<mxid>[^/]*)$")
 
-    # @override
     async def on_GET(self, request: SynapseRequest, mxid: str) -> tuple[int, JsonDict]:
         requester = await self.api.get_user_by_req(request)
+        await self.permission_handler.do_migration(self.store, requester.user)
```
nico-famedly, line 129: So if I am reading this correctly, we do the migration individually for each user only when they use the contact API. I don't think we have that many users on existing servers, that migrating all old users once as a background job on startup is a problem. I think that would reduce a lot of complexity and with the current approach it is very hard to tell, what state we are in? So I think we should just migrate and drop a lot of the database code.

jason-famedly: I was unaware what our perspective user base looked like and planned for the likelihood that it would be several thousand. If you think it'll be fine without it, I'll remove the on-demand migration

synapse_invite_checker/rest/contacts.py
jason-famedly: I had made the assumption that the client app would be allowed time to roll out some of the changes and had considered the likeliest option was migration first and *then* allowing that to change. From the moment of migration the data is available to the client and it shouldn't need to use the contact API at all

synapse_invite_checker/invite_checker.py
jason-famedly: It was the option I was able to find to hide that it wouldn't show on the coverage report. It didn't appear to work anyway. It can go

README.md
jason-famedly: I thought the name was rather self explanatory. I can change it to `force-contact-migration-to-permissions` if you would rather?

synapse_invite_checker/rest/contacts.py
jason-famedly: What error code do you think would be appropriate? 403?

jason-famedly:
> we need to return an error if the account data was directly modified in ways that conflict with the old contact API

I got you, some how I must have misunderstood that. Should be more in line with that now

synapse_invite_checker/rest/contacts.py
jason-famedly: removed

synapse_invite_checker/invite_checker.py
jason-famedly: removed in [39778bb](https://github.com/famedly/synapse-invite-checker/pull/40/commits/39778bb5a34f586f8766bece0a55495d7835b9fb)

README.md
jason-famedly: Actually, I just removed it altogether. It's safe to run on each server restart, as if there is nothing to migrate it just will not do anything

Removed in [42e465f](https://github.com/famedly/synapse-invite-checker/pull/40/commits/42e465fe6e5dbcc7a7d9e02c04605fb3299ee3da)

nico-famedly (approved):

PR #41, opened 2025-02-03
Check and limit room version when creating a room

jason-famedly:
This will add a check into the `on_create_room()` for verifying that the room version requested is allowed.

New configuration setting: `allowed_room_versions` which needs to be set as a list

```
allowed_room_versions:
- "9"
- "10"
```

helps with 
* famedly/product-management#2824

### DIFF
--- a/synapse_invite_checker/config.py
+++ b/synapse_invite_checker/config.py
@@ -12,7 +12,7 @@
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from synapse_invite_checker.types import TimType
 
@@ -27,3 +27,4 @@ class InviteCheckerConfig:
     federation_localization_url: str = ""
     gematik_ca_baseurl: str = ""
     tim_type: TimType = TimType.PRO
+    allowed_room_versions: list[str] = field(default_factory=list)
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -269,6 +269,16 @@ class InviteChecker:
                 msg = "`tim-type` setting is not a recognized value. Please fix."
                 raise ConfigError(msg)
 
+        _allowed_room_versions = config.get("allowed_room_versions", ["9", "10"])
+        if not _allowed_room_versions or not isinstance(_allowed_room_versions, list):
+            msg = "Allowed room versions must be formatted as a list."
+            raise ConfigError(msg)
+
+        _config.allowed_room_versions = [
+            # Coercing into a string, in case the yaml loader thought it was an int
+            str(_room_ver)
+            for _room_ver in _allowed_room_versions
+        ]
         return _config
 
     def after_startup(self) -> None:
@@ -463,7 +473,9 @@ class InviteChecker:
         is_request_admin: bool,
     ) -> None:
         """
-        Raise a SynapseError if creating a room with invites should be denied
+        Raise a SynapseError if creating a room should be denied. Currently, this checks
+        invites
+        room version
         """
         # Unlike `user_may_invite()`, `on_create_room()` only runs with the inviter being
         # a local user and the invitee is remote. Unfortunately, the spam check module
@@ -492,6 +504,20 @@ class InviteChecker:
                     f"Room not created as user ({invitee}) is not allowed to be invited",
                     errors.Codes.FORBIDDEN,
                 )
+
+        # The room version should always be a string to accommodate arbitrary unstable
+        # room versions. If it was not explicitly requested, the homeserver defaults
+        # will be used. Make sure to check that instance as well
+        room_version: str = request_content.get(
+            "room_version", self.api._hs.config.server.default_room_version.identifier
+        )
+
+        if room_version not in self.config.allowed_room_versions:
+            raise SynapseError(
+                400,
+                f"Room version ('{room_version}') not allowed",
+                errors.Codes.FORBIDDEN,
+            )
 
     async def user_may_invite(
         self, inviter: str, invitee: str, room_id: str | None = None
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@ modules:
         federation_list_client_cert: "tests/certs/client.pem", # path to a pem encoded client certificate for mtls, required if federation list url is https
         gematik_ca_baseurl: "https://download-ref.tsl.ti-dienste.de/", # the baseurl to the ca to use for the federation list, required
         tim-type: "epa" or "pro", # Patient/Insurance or Professional mode, defaults to "pro" mode. Optional currently, but will be required in a later release
+        allowed_room_versions: # The list(as strings) of allowed room versions. Currently optional, defaults are listed
+          - "9"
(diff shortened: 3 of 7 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -257,6 +257,23 @@ def parse_config(config: Dict[str, Any]) -> InviteCheckerConfig:
...
+        _allowed_room_versions = config.get("allowed_room_versions", [])
+        if not _allowed_room_versions or not isinstance(_allowed_room_versions, list):
+            msg = "Allowed room versions must be formatted as a list. Please fix"
+            raise ConfigError(msg)
+
+        try:
+            _config.allowed_room_versions.extend(
+                # By coercing into first an integer, and then a string, we guarantee
+                # it is a number
+                str(int(_room_ver))
```
nico-famedly, line 269: room versions are not numbers.

nico-famedly:
Your example in the description looks wrong? I would expect strings and a list with multiple entries, i.e.

```
- "9"
- "10"
```

instead of 

```
- 9
   10
```

synapse_invite_checker/config.py
```diff
@@ -27,3 +27,4 @@ class InviteCheckerConfig:
     federation_localization_url: str = ""
     gematik_ca_baseurl: str = ""
     tim_type: TimType = TimType.PRO
+    allowed_room_versions: list = field(default_factory=list)
```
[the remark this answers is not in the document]
jason-famedly: While in draft, it's a secret :shushing_face:

synapse_invite_checker/invite_checker.py
jason-famedly: That's right. I did want to enforce that the string passed in was actually a number. I guess that part isn't strictly necessary. We will have custom room versions later on, I suppose?

synapse_invite_checker/config.py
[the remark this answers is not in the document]
jason-famedly: added to README in [138ffc5](https://github.com/famedly/synapse-invite-checker/pull/41/commits/138ffc57236bd81e20921367d8919785d65f9c6b)

synapse_invite_checker/invite_checker.py
jason-famedly: I did go ahead and pull this out. It will just have to accept the string, whatever that may be

synapse_invite_checker/config.py
nico-famedly, line 30: How is this different from this?

```suggestion
    allowed_room_versions: list[str] = []
```

synapse_invite_checker/invite_checker.py
```diff
@@ -269,6 +269,16 @@ def parse_config(config: Dict[str, Any]) -> InviteCheckerConfig:
                 msg = "`tim-type` setting is not a recognized value. Please fix."
                 raise ConfigError(msg)
 
+        _allowed_room_versions = config.get("allowed_room_versions", ["9", "10"])
+        if not _allowed_room_versions or not isinstance(_allowed_room_versions, list):
+            msg = "Allowed room versions must be formatted as a list. Please fix"
```
nico-famedly, line 274:
```suggestion
            msg = "Allowed room versions must be formatted as a list."
```

synapse_invite_checker/invite_checker.py
```diff
@@ -269,6 +269,16 @@ def parse_config(config: Dict[str, Any]) -> InviteCheckerConfig:
...
+            raise ConfigError(msg)
+
+        _config.allowed_room_versions.extend(
+            # Coercing into a string, in case the yaml loader thought it was an int
+            str(_room_ver)
+            for _room_ver in _allowed_room_versions
+        )
```
nico-famedly, line 281:
```suggestion
        _config.allowed_room_versions = [
            # Coercing into a string, in case the yaml loader thought it was an int
            str(_room_ver)
            for _room_ver in _allowed_room_versions
        ]
```

Any reason you used `extend` there?

synapse_invite_checker/invite_checker.py
```diff
@@ -493,6 +505,16 @@ async def on_create_room(
                     errors.Codes.FORBIDDEN,
                 )
 
+        # The room version should always be a string to accommodate arbitrary unstable room versions
+        room_version: str | None = request_content.get("room_version", None)
+
+        if room_version and room_version not in self.config.allowed_room_versions:
```
nico-famedly, line 511: We should also error if the default room version is not allowed here.

synapse_invite_checker/config.py
jason-famedly: From what I understand, setting a class variable with a mutable object(which is what `[]` is in this case, I believe it's called a literal?) contaminates any use of `[]` later. Same reason we can't use `[]` as a default for kwargs, it is bad. So for dataclasses, we use the `field()` with `default_factory=list` to cause the generated dunder method of `__init__()` generate a fresh empty `list` for us! :grin: 

(This one burned me a few years ago, hence my being conscious of it now)

synapse_invite_checker/invite_checker.py
jason-famedly: I think I was setting the defaults a different way in a previous iteration. Since the empty list is already generated, this seemed more prudent than recreating the list. But I'm not married to it, this will be better

synapse_invite_checker/invite_checker.py
jason-famedly: Good shout, forgot about that already

synapse_invite_checker/invite_checker.py
jason-famedly: fixed in [6948be3](https://github.com/famedly/synapse-invite-checker/pull/41/commits/6948be37766665fe0b984b07358fed77be35d375)

synapse_invite_checker/invite_checker.py
jason-famedly: fixed in [6948be3](https://github.com/famedly/synapse-invite-checker/pull/41/commits/6948be37766665fe0b984b07358fed77be35d375)

synapse_invite_checker/config.py
nico-famedly: Makes sense.

synapse_invite_checker/invite_checker.py
```diff
@@ -493,6 +505,20 @@ async def on_create_room(
...
                 )
 
+        # The room version should always be a string to accommodate arbitrary unstable
+        # room versions. If it was not explicitly requested, the homeserver defaults
+        # will be used. Make sure to check that instance as well
+        room_version: str = request_content.get(
+            "room_version", self.api._hs.config.server.default_room_version.identifier
+        )
+
+        if room_version and room_version not in self.config.allowed_room_versions:
```
nico-famedly, line 515:
```suggestion
        if room_version not in self.config.allowed_room_versions:
```

You don't want to allow empty strings as the room version, right?

jason-famedly: Good catch, I thought I had removed that

fixed in [c34810a](https://github.com/famedly/synapse-invite-checker/pull/41/commits/c34810ae73ecffde5ad73bdc5c93eb06ab4457f7)

nico-famedly (approved):

PR #43, opened 2025-02-10
Remove rooms that only have EPA members left in them

jason-famedly:
My current take on this is to use the `TaskScheduler` provided by Synapse to schedule purging rooms just as is done by the Admin API and the "room forgetting" options. This allows the worker(or main process if it isn't defined) nominated by the Synapse setting `run_background_tasks_on` to handle both, and allows for resuming these tasks after a server restart.

Running a room scan will be on a `looping_background_call()` schedule so as not to clash with any room deletions currently in progress

The `TaskScheduler` does have a few things to be aware of:
* Has a 1 minute window between checks for new work, if it is not currently active
* Allows a maximum of 5 `ScheduledTask`s to run at the same time(despite what it's docstring says)
* Resumes previously scheduled work after a server restart

A setting section and 3 new settings are introduced:
* `room_scan_run_interval`: To determine how often a room scan should be performed. Defaults to 1 hour with an enforced minimum of 1 hour, setting to `0` completely turns off the room scanner. 
* `insured_only_room_scan`: contains the two below
   * `enabled`: True by default if the server is in EPA mode otherwise False, allows for turning off the insured-only room scan. Forced to False when server is in PRO mode.
   * `grace_period`: To determine how long a room may survive after all PRO users have left. Defaults to 1 week, ignored if `enabled` is False

These new settings use a "shorthand" method of defining the required milliseconds(however, milliseconds may also be used if preferred). See the README file for more information.

Two new metrics to watch for:
* For "Per-block metrics" section, "get_timestamp_from_eligible_events_for_epa_room_purge" will be found on the Synapse worker for background task handling
* For "Background Jobs" section, "room_scan" should now be found, again on the Synapse worker for background tasks

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,15 @@ modules:
         allowed_room_versions: # The list(as strings) of allowed room versions. Currently optional, defaults are listed
           - "9"
           - "10"
+        room_scan_run_interval: see 'Duration Parsing' below, # How often to scan for rooms that are eligible for deletion. Defaults to "1h". Setting to "0" completely disables all room scanning
+        insured_only_room_scan:
+          enabled: true or false  # optional switch to disable the insured-only room scan from running.  The scan is enabled by default, but only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
+          grace_period: see 'Duration Parsing' below, # Length of time a room with only EPA members is allowed to exist before deletion. Ignored if `enabled` is false. Defaults to "1w"
 ```
-
+### Duration Parsing
+Settings labeled as 'duration_parsing' allow for a string representation of the value
+that is converted to milliseconds. Suffixes with 's', 'm', 'h', 'd', 'w', or 'y' may be used. For example:
+`1h` would translate to `3600000` milliseconds
 ## Testing
 
 The tests uses twisted's testing framework trial, with the development
--- a/synapse_invite_checker/config.py
+++ b/synapse_invite_checker/config.py
@@ -18,6 +18,12 @@ from synapse_invite_checker.types import TimType
 
 
 @dataclass
+class InsuredOnlyRoomScanConfig:
+    grace_period_ms: int = 0
+    enabled: bool = False
+
+
+@dataclass
 class InviteCheckerConfig:
     title: str = "Invite Checker module by Famedly"
     description: str = "Invite Checker module by Famedly"
@@ -28,3 +34,7 @@ class InviteCheckerConfig:
     gematik_ca_baseurl: str = ""
     tim_type: TimType = TimType.PRO
     allowed_room_versions: list[str] = field(default_factory=list)
+    room_scan_run_interval_ms: int = 0
+    insured_room_scan_options: InsuredOnlyRoomScanConfig = field(
+        default_factory=InsuredOnlyRoomScanConfig
+    )
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -16,7 +16,7 @@ import base64
 import functools
 import logging
 from contextlib import suppress
-from typing import Any, Callable, Dict, Literal, cast
+from typing import Any, Callable, Dict, Literal
 from urllib.parse import quote, urlparse
 
 from cachetools import TTLCache, keys
@@ -30,17 +30,23 @@ from OpenSSL.crypto import (
     dump_certificate,
     load_certificate,
 )
-from synapse.api.constants import AccountDataTypes
+from synapse.api.constants import AccountDataTypes, EventTypes, Membership
 from synapse.api.errors import SynapseError
 from synapse.api.room_versions import RoomVersion
 from synapse.config import ConfigError
+from synapse.config._base import Config
+from synapse.events import EventBase
+from synapse.handlers.pagination import SHUTDOWN_AND_PURGE_ROOM_ACTION_NAME
 from synapse.http.client import BaseHttpClient
 from synapse.http.proxyagent import ProxyAgent
 from synapse.http.server import JsonResource
 from synapse.module_api import NOT_SPAM, ModuleApi, errors
 from synapse.server import HomeServer
-from synapse.storage.database import make_conn
-from synapse.types import Requester, UserID
+from synapse.storage.database import LoggingTransaction, make_conn
+from synapse.types import Requester, ScheduledTask, TaskStatus, UserID
+from synapse.types.handlers import ShutdownRoomParams
+from synapse.types.state import StateFilter
+from synapse.util.metrics import measure_func
 from twisted.internet.defer import Deferred
 from twisted.internet.ssl import PrivateCertificate, optionsForClientTLS, platformTrust
 from twisted.web.client import HTTPConnectionPool
@@ -159,7 +165,8 @@ class InviteChecker:
 
     def __init__(self, config: InviteCheckerConfig, api: ModuleApi):
         self.api = api
-
+        # Need this for the @measure_func decorator to work
+        self.clock = api._hs.get_clock()
         self.config = config
 
         self.federation_list_client = FederationAllowListClient(api._hs, self.config)
@@ -193,6 +200,18 @@ class InviteChecker:
             self.fetch_localization_for_mxid,
             self.is_domain_insurance,
         )
+
+        self.task_scheduler = api._hs.get_task_scheduler()
+
+        if self.config.room_scan_run_interval_ms > 0:
+            # The docstring for 'looping_background_call()' is slightly incorrect
+            # > Waits msec initially before calling f for the first time.
+            # Should be
+            # > Calls f after waiting msec, then repeats. This is an inexact, "best effort"
+            # > figure when the reactor/event loop is under heavy load
+            self.api.looping_background_call(
+                self.room_scan, self.config.room_scan_run_interval_ms
+            )
 
         # The Contact Management API resources
         ContactManagementInfoResource(self.config).register(self.resource)
@@ -283,16 +302,49 @@ class InviteChecker:
             str(_room_ver)
             for _room_ver in _allowed_room_versions
         ]
+
+        run_interval = Config.parse_duration(config.get("room_scan_run_interval", "1h"))
+        clamp_minimum_to = Config.parse_duration("1h")
+
+        # If 'room_scan_run_interval' is not set to 0 for disabling the room scan
+        # completely, make sure anything less than the minimum of 1 hour is ignored
+        _config.room_scan_run_interval_ms = (
+            max(run_interval, clamp_minimum_to) if run_interval > 0 else run_interval
+        )
+
+        insured_room_scan_section = config.get("insured_only_room_scan", {})
+        if not isinstance(insured_room_scan_section, dict):
+            msg = "`insured_only_room_scan` should be configured as a dictionary"
+            raise ConfigError(msg)
+
+        # Only default enable this room scan if in EPA mode
+        enable_insured_room_scan = insured_room_scan_section.get(
+            "enabled", True if _config.tim_type == TimType.EPA else False
+        )
+        # But also prevent it running in PRO mode completely
+        enable_insured_room_scan = (
+            False if _config.tim_type == TimType.PRO else enable_insured_room_scan
+        )
+        _config.insured_room_scan_options.enabled = enable_insured_room_scan
+
+        epa_room_grace_period = Config.parse_duration(
+            insured_room_scan_section.get("grace_period", "1w")
+        )
+
+        _config.insured_room_scan_options.grace_period_ms = epa_room_grace_period
+
         return _config
 
     def after_startup(self) -> None:
+        _ = Deferred.fromCoroutine(self._after_startup())
+
+    async def _after_startup(self) -> None:
         """
         To be called when the reactor is running. Validates that the epa setting matches
         the insurance setting in the federation list and *might* perform forced contact
         migration.
         """
-        fetch_deferred = Deferred.fromCoroutine(self._fetch_federation_list())
-        fed_list = cast(FederationList, fetch_deferred.result)
+        fed_list = await self._fetch_federation_list()
         if self.config.tim_type == TimType.EPA and not fed_list.is_insurance(
             self.api._hs.config.server.server_name
         ):
@@ -301,7 +353,7 @@ class InviteChecker:
                 "the Federation List as an Insurance Domain!"
             )
 
-        _ = Deferred.fromCoroutine(self.run_migration())
+        await self.run_migration()
 
     async def run_migration(self) -> None:
         """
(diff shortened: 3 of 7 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -604,3 +643,162 @@ async def user_may_invite(
...
+
+    async def get_all_room_ids(self) -> list[str]:
+        """Retrieve all room IDS."""
+
+        # There is an PRIMARY index on room_id
+        def f(txn: LoggingTransaction) -> list[str]:
+            sql = "SELECT room_id FROM rooms"
+            txn.execute(sql)
+            result = [room_id for (room_id,) in txn.fetchall()]
+            return result
```
jason-famedly, line 655: The EXPLAIN for my server with 410 rooms looked like:
```
synapse=# EXPLAIN SELECT room_id FROM rooms;
                       QUERY PLAN                        
---------------------------------------------------------
 Seq Scan on rooms  (cost=0.00..10.10 rows=410 width=32)
(1 row)
```

synapse_invite_checker/invite_checker.py
```diff
@@ -604,3 +643,162 @@ async def user_may_invite(
...
+        #  (room_id, type, state_key) (unique)
+        #  state_key
+        #  (room_id, membership)
+        # events has an index on event_id
+        def f(txn: LoggingTransaction) -> list[tuple[str, str, int]]:
+            sql = """
+            SELECT c.state_key, c.membership, e.received_ts
+            FROM current_state_events as c
+            INNER JOIN events e ON c.event_id=e.event_id
+            WHERE c.room_id = '%s' AND c.type = 'm.room.member'
```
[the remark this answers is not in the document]
jason-famedly: (Moved from source comment):
```
current_state_events has an indexes on:
event_id(unique)
(room_id, type, state_key) (unique)
state_key
(room_id, membership)
events has an index on event_id
```

jason-famedly:
Resolves famedly/product-management#2828

synapse_invite_checker/invite_checker.py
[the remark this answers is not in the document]
jason-famedly: This no longer exists, ignore

synapse_invite_checker/invite_checker.py
```diff
@@ -642,3 +680,125 @@ async def user_may_invite(
...
+            room_id,
+            delete_id,
+        )
+
+    async def room_scan(self) -> None:
+        """
+        Scan all rooms for eligible conditions to shutdown and purge a room.
+        """
+        if self.config.tim_type == TimType.PRO:
+            return
```
jason-famedly, line 778: This will end up being removed when we start purging inactive rooms

README.md
```diff
@@ -41,8 +41,14 @@ modules:
         allowed_room_versions: # The list(as strings) of allowed room versions. Currently optional, defaults are listed
           - "9"
           - "10"
+        room_scan_run_interval: see 'Duration Parsing' below, # How often to scan for rooms that are eligible for deletion. Defaults to "1h"
+        epa_room_survival_duration: see 'Duration Parsing' below, # Length of time a room with only EPA members is allowed to exist before deletion. Defaults to "1w"
+        enable_room_scan: true or false  # optional switch to disable the room scan from running, defaults to true
```
nico-famedly, line 46: Is this the epa room scan or the inactive room scan?

synapse_invite_checker/invite_checker.py
```diff
@@ -642,3 +675,124 @@ async def user_may_invite(
...
+                    await self.get_last_pro_users_leave_timestamp(room_id)
+                )
+
+                # last_user_left_timestamp being None is our indicator that no PRO users
+                # were ever in this room. There may still be a pending invite so ignore it
+                if last_user_left_timestamp and (
+                    last_user_left_timestamp + self.config.epa_room_survival_duration_ms
+                    <= self.api._hs.get_clock().time_msec()
+                ):
+                    rooms_to_purge.add(room_id)
```
nico-famedly, line 795: I think we should also purge rooms, where we never had a pro user in the room? Not sure, how we could get to that case, but that seems sensible?

synapse_invite_checker/invite_checker.py
```diff
@@ -642,3 +675,124 @@ async def user_may_invite(
...
+        If the room has already been scheduled or is actively being deleted, do nothing.
+        If the room was purged already in the past, but for some reason is still hanging
+        around in the database, try it again(I know this feels wrong, but ask me how I
+        know this is a problem worth accounting for)
+        """
+        if len(await self.get_delete_tasks_by_room(room_id)) > 0:
+            logger.warning("Purge already in progress or scheduled for %s" % (room_id,))
+            return
+
+        shutdown_params = {
```
nico-famedly, line 745: This? https://github.com/famedly/synapse/blob/b5c06e962b3e92b67d0ae6a19b94d1ea311507ee/synapse/types/handlers/__init__.py#L39

README.md
jason-famedly: This is for the epa room scan. Oh, I see why you are asking, yeah I'll update all the references so there will be two

synapse_invite_checker/invite_checker.py
jason-famedly: This was to account for pending invites that are delayed, so the PRO user has time to accept the invite(could be hours or days later). There was an undefined period for such, and I considered maybe having the inactive rooms scan pick it up?

synapse_invite_checker/invite_checker.py
jason-famedly: Yes `ShutdownRoomParams` is a `TypedDict`. So, it's just a regular Dict with extra junk attached. Since the values for the shutting down a room aren't variable, don't see a need to create that exact type of object. If you disagree, I'll make it do that instead. This is how they do it on the Rest endpoint as inside it's just a `Mapping` and there are no defaults on instantiation of the object anyway

synapse_invite_checker/invite_checker.py
jason-famedly: superceded by later work

README.md
jason-famedly: This was adjusted based on out-of-band communication for insight and clarity. 

See [14aa79a](https://github.com/famedly/synapse-invite-checker/pull/43/commits/14aa79a81d7ab63f8004f98e7467942b39b1bd2f)

synapse_invite_checker/invite_checker.py
jason-famedly: Fixed in [14aa79a](https://github.com/famedly/synapse-invite-checker/pull/43/commits/14aa79a81d7ab63f8004f98e7467942b39b1bd2f)

synapse_invite_checker/invite_checker.py
jason-famedly: Fixed in [14aa79a](https://github.com/famedly/synapse-invite-checker/pull/43/commits/14aa79a81d7ab63f8004f98e7467942b39b1bd2f)

synapse_invite_checker/config.py
```diff
@@ -17,6 +17,12 @@
 from synapse_invite_checker.types import TimType
 
 
+@dataclass
+class InsuredOnlyRoomScanConfig:
+    epa_room_grace_period_ms: int = 0
```
nico-famedly, line 22: Not sure if the epa makes that much sense in the name, but I guess with the current spec it does.
Also maybe this shouldn't default to a grace period of 0 and should default to enabled?

synapse_invite_checker/invite_checker.py
```diff
@@ -283,16 +303,37 @@ def parse_config(config: Dict[str, Any]) -> InviteCheckerConfig:
             str(_room_ver)
             for _room_ver in _allowed_room_versions
         ]
+
+        run_interval = Config.parse_duration(config.get("room_scan_run_interval", "1h"))
+        _config.room_scan_run_interval_ms = run_interval
```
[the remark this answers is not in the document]
nico-famedly: Yep, sounds good

synapse_invite_checker/invite_checker.py
```diff
@@ -642,3 +684,143 @@ async def user_may_invite(
...
+
+    @measure_func("get_last_pro_users_leave_timestamp")
+    async def get_last_pro_users_leave_timestamp(self, room_id) -> int | None:
+        """
+        Retrieve and parse the room PRO members that left into a timestamp of the
+        latest event, or None if there were no PRO members in the room
+        """
+        state_mapping: dict[tuple[str, str], EventBase] = (
+            await self.api._storage_controllers.state.get_current_state(
+                room_id, StateFilter.from_types([(EventTypes.Member, None)])
```
nico-famedly, line 708: Should we just also fetch the create event here and default to that if nothing else is found?

tests/test_scheduled_tasks.py
```diff
@@ -0,0 +1,476 @@
...
+
+        # wait for cleanup, should take 6 hours(based on above configuration)
+        count = 0
+        while True:
+            count += 1
+            if count == 6:
+                break
+
+            # advance() is in seconds, this should be 1 hour
+            self.reactor.advance(60 * 60)
```
nico-famedly, line 202: I'm a bit scared this might flake if you advance by exactly 6h? Maybe you should pick like 59 minutes per advance? :D

nico-famedly (commented):
Apart from a few nitpicks, this seems fine in general

synapse_invite_checker/config.py
jason-famedly: It should only default to enabled on epa servers, I think. Provided that Synapse's `forget_rooms_on_leave` and `forgotten_room_retention_period` are set, it serves no purpose to run on pro servers?

But that did make me question how the defaults are implemented, I'll go take another look at that

tests/test_scheduled_tasks.py
jason-famedly: We take advantage of the `TaskScheduler` having a resolution of 1 minute to absorb that. It seemed like it to me too. The scheduling is all that happens here. I used to test that exact condition(and even when testing with postgres which runs everything much slower) and it always kept turning out just fine. I did remove the scheduling part of the test as I felt it was redundant. I think it'll be fine, if it does end up flaking later one or shows a sign we can adjust it. (Also take note that this loop only does 5 hours, and the last hour is done after the break. This is a remnant of the assumption I needed to do that 59 minutes for the last bit. It turned out to not be necessary but I left it in anyway. I should probably comment something to that effect)

synapse_invite_checker/invite_checker.py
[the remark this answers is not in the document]
jason-famedly: What do you think the minimum should be? I'm leaning towards 1 hour, in case this does end up being millions of users with hundreds of thousands of rooms to scan

synapse_invite_checker/config.py
jason-famedly: Focusing on this the grace period defaulting to `0` here actually seems ok. And here's why I think that:
* I did change out how the config is parsed in `parse_config()` so that it always has at the minimum an empty `Dict` to look at(and not None anymore)
* Since that logic no longer hides it all behind checking for the None, any defaults defined in `parse_config()` should now work unilaterally
* In the chance of a `ConfigError` raising(something I just added), all of this is moot and the `0` would be ignored and the room scan would be disabled anyway because of the `enabled` also being False

Does that make sense?

synapse_invite_checker/invite_checker.py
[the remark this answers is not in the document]
jason-famedly: I set it to 1 hour, just to move along in [916b94b](https://github.com/famedly/synapse-invite-checker/pull/43/commits/916b94b162374ebd31e7789640b8193042581beb)

synapse_invite_checker/config.py
jason-famedly: The rest of the adjustments here were done in [916b94b](https://github.com/famedly/synapse-invite-checker/pull/43/commits/916b94b162374ebd31e7789640b8193042581beb)

synapse_invite_checker/invite_checker.py
jason-famedly: Should be fixed up in [916b94b](https://github.com/famedly/synapse-invite-checker/pull/43/commits/916b94b162374ebd31e7789640b8193042581beb)

synapse_invite_checker/invite_checker.py
```diff
@@ -642,3 +684,143 @@ async def user_may_invite(
...
+        all_room_ids = await self.get_all_room_ids()
+
+        logger.debug("Detected %d total rooms", len(all_room_ids))
+
+        rooms_to_purge = set()
+        for room_id in all_room_ids:
+            hosts = await self.api._store.get_current_hosts_in_room(room_id)
+            epa_hosts = {host for host in hosts if await self.is_domain_insurance(host)}
+            # only purge rooms that only have EPA hosts in them
+            if hosts == epa_hosts:
```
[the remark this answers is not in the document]
jason-famedly: done in [916b94b](https://github.com/famedly/synapse-invite-checker/pull/43/commits/916b94b162374ebd31e7789640b8193042581beb)

synapse_invite_checker/config.py
nico-famedly: Yeah, that's fine. It just seemed weird, because I read the diff from the top.

And yes, it only makes sense to run this on pro servers, I just think the "epa" makes little sense to include in the config key, since the whole check is epa only. I think the check should just silently do nothing when enabled on pro servers.

tests/test_scheduled_tasks.py
nico-famedly: Hm, okay, if you say so...

synapse_invite_checker/config.py
jason-famedly: See what you think about it now in [d515637](https://github.com/famedly/synapse-invite-checker/pull/43/commits/d515637e7340e54713809a5899f5209512773bb9)

nico-famedly (dismissed):

README.md
```diff
@@ -41,8 +41,15 @@ modules:
         allowed_room_versions: # The list(as strings) of allowed room versions. Currently optional, defaults are listed
           - "9"
           - "10"
+        room_scan_run_interval: see 'Duration Parsing' below, # How often to scan for rooms that are eligible for deletion. Defaults to "1h". Setting to "0" completely disables all room scanning
+        insured_only_room_scan:
+          enabled: true or false  # optional switch to disable the insured-only room scan from running, defaults to true based on 'tim-type: "epa"' setting above
```
nico-famedly, line 46: Comment is wrong, should say that it can only be enabled in epa mode?

jason-famedly:
```suggestion
          enabled: true or false  # optional switch to disable the insured-only room scan from running. If 'tim-type` above is set to "epa" this defaults to true, if above is set to "pro" this is forced to false
```
How's that sound?

nico-famedly: How about just:

```suggestion
          enabled: true or false  # optional switch to disable the insured-only room scan from running. The scan only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
```

jason-famedly: done it [2e6a196](https://github.com/famedly/synapse-invite-checker/pull/43/commits/2e6a196e12b52f980ec03a5f2789ee24cb0661e8)

nico-famedly (approved):

PR #45, opened 2025-02-20
Scan and purge inactive rooms

jason-famedly:
Resolves famedly/product-management#2829 

Find and remove inactive rooms. Do this by finding the last actual message in a room and retrieving it's time stamp. In the case that the room never had any messages(either encrypted or clear text), use the room creation event. If it was longer than the `grace_period` below, that room is scheduled for purging using the infrastructure introduced in #43 . 

Frequency of the scan itself is governed by `room_scan_run_interval`

New settings
```yaml
inactive_room_scan:
  enabled: true or false # an optional switch to disable processing of this part of the room scan
  grace_period: 26w # An optional value for how long to keep a room after the last eligible event,
                    # defaults to 6 months("26w"). Uses the 'duration parsing' or a direct millisecond
                    # integer. Ignored if 'enabled' is false
```

New metric available for "Per-block Metrics" section:
* "get_timestamp_of_last_eligible_activity_in_room" will be found on the Synapse worker responsible for background tasks

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -30,8 +30,9 @@ from OpenSSL.crypto import (
     dump_certificate,
     load_certificate,
 )
-from synapse.api.constants import AccountDataTypes, EventTypes, Membership
+from synapse.api.constants import AccountDataTypes, Direction, EventTypes, Membership
 from synapse.api.errors import SynapseError
+from synapse.api.filtering import Filter
 from synapse.api.room_versions import RoomVersion
 from synapse.config import ConfigError
 from synapse.config._base import Config
@@ -333,6 +334,22 @@ class InviteChecker:
 
         _config.insured_room_scan_options.grace_period_ms = epa_room_grace_period
 
+        # This option is considered for all server modes unlike 'insured_only_room_scan'
+        inactive_room_scan_section = config.get("inactive_room_scan", {})
+        if not isinstance(inactive_room_scan_section, dict):
+            msg = "`inactive_room_scan` should be formatted as a dictionary"
+            raise ConfigError(msg)
+
+        enable_inactive_room_scan = inactive_room_scan_section.get("enabled", True)
+        _config.inactive_room_scan_options.enabled = enable_inactive_room_scan
+
+        # "26w" calculates as 6 months
+        inactive_room_scan_grace_period = Config.parse_duration(
+            inactive_room_scan_section.get("grace_period", "26w")
+        )
+        _config.inactive_room_scan_options.grace_period_ms = (
+            inactive_room_scan_grace_period
+        )
         return _config
 
     def after_startup(self) -> None:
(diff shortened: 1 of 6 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -742,6 +760,51 @@ async def get_timestamp_from_eligible_events_for_epa_room_purge(
...
+        page_config = PaginationConfig(
+            None,
+            None,
+            direction=Direction.BACKWARDS,
+            # Not sure what to use for a limit here, there may have been 10's of events.
+            # I will try and filter them so this number can be smol
+            limit=5,
+        )
+        # This is apparently a RoomEventFilter
+        filter_json = {"types": [EventTypes.Message, EventTypes.Encrypted]}
```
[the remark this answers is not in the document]
nico-famedly: (Or invert the filter and only filter out member events)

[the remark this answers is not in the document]
jason-famedly: I will write a test case for that behavior. 

Perhaps look at create event as a second call if no messages were found. I don't know about inverting the filter, that would pull a bunch of other stuff in to and I thought we didn't want to be that broad with the scope of where the timestamp comes from?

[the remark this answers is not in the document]
jason-famedly: Included in Final draft

synapse_invite_checker/invite_checker.py
```diff
@@ -742,6 +759,51 @@ async def get_timestamp_from_eligible_events_for_epa_room_purge(
...
+            to_key=None,
+            direction=Direction.BACKWARDS,
+            # With the filter below, 5 should be more than enough
+            limit=5,
+            event_filter=event_filter,
+        )
+        # If we succeeded with event type filtering, all of these should be only messages
+        last_timestamp = 0
+        for event_base in events:
+            last_timestamp = max(event_base.origin_server_ts, last_timestamp)
```
nico-famedly, line 803: Would this cause backfill to delete the room? Or does topological ordering prevent that? Not something to fix on this PR, I am just interested in your thoughts.

tests/test_room_message_timestamp.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_can_find_last_message_timestamp(self) -> None:
+        # self.hs.mockmod: InviteChecker
+        # create a room, add another user
+        # send a message, get the timestamp
+        # send two more messages, get the timestamp
+        # have other user send a message, get that timestamp
+        # have other user send two more messages, and get that timestamp
+
+        def send_message_and_get_timestamp(room, message, tok) -> None:
```
nico-famedly, line 68:
```suggestion
        def send_message_and_assert_latest_activity(room, message, tok) -> None:
```

It took me a bit to understand what this is for, since it never returns the timestamp, that it "gets". Nitpick, but mybe this is clearer?

nico-famedly (approved):

synapse_invite_checker/invite_checker.py
jason-famedly: NOTES for future us: I think this could only happen if the events coming in were at the newest tip of the DAG but also had a timestamp older than the `grace_period`.  So, possibly but rarely/unlikely. Topological ordering seems to be synonymous with 'depth' in a lot of cases

tests/test_room_message_timestamp.py
jason-famedly: Yes, good catch

jason-famedly:
Please check the wording on the commit itself

nico-famedly (approved):

PR #55, opened 2025-02-26
chore: Disable mounting Contacts API endpoints when in EPA mode

jason-famedly:
There is no point in even having these endpoints live when in EPA mode, as no Contacts should exist on EPA servers

depends on #60

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -227,18 +227,18 @@ class InviteChecker:
         # Separate out the resources for Contacts, since they will be going away
         self.contact_resource = JsonResource(api._hs)
 
-        # The Contact Management API resources
-        ContactManagementInfoResource(self.config).register(self.contact_resource)
-        ContactsResource(self.api, self.store, self.permissions_handler).register(
-            self.contact_resource
-        )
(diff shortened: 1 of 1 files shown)


jason-famedly:
Took advantage of being near the remote invite tests to break them into finer and simpler pieces

tests/test_remote_invites.py
```diff
@@ -327,51 +254,46 @@ async def may_invite(self, inviter: str, invitee: str, roomid: str):
             return NOT_SPAM
         return ret[0]  # return first code instead of all of them to make assert easier
 
-    async def test_invite_from_remote_not_on_fed_list(self) -> None:
+    def test_invite_from_remote_not_on_fed_list(self) -> None:
         """Tests that an invite from a remote server not in the federation list gets denied"""
-        self.add_a_contact_to_user_by_token(
```
nico-famedly, line 332: Why did you remove this? Shouldn't you be able to just use the account data event here?

jason-famedly:
Separated the test refactor out into a separate branch

nico-famedly (approved):
Seems fine, I am just a bit confused by the coverage not updating...

PR #58, opened 2025-03-04
feat: Forbid local user from joining remote rooms(with no invite)

jason-famedly:
part of famedly/product-management#2826
Requires #55 

Public rooms do not require invites to join, unlike private rooms. Take advantage of this to block joining a room not on the local server if an invite does not exist

Prohibiting a join when there is an invite will be a separate pull request

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -19,7 +19,10 @@ from http import HTTPStatus
 from typing import Any
 from unittest.mock import AsyncMock, Mock, patch
 
+from synapse.api.constants import Membership, RoomCreationPreset
+from synapse.api.errors import SynapseError
 from synapse.api.room_versions import KNOWN_ROOM_VERSIONS
+from synapse.crypto.event_signing import add_hashes_and_signatures
 from synapse.events import make_event_from_dict
 from synapse.module_api import errors
 from synapse.rest import admin
@@ -33,19 +36,21 @@ from synapse.rest.client import (
     room_upgrade_rest_servlet,
 )
 from synapse.server import HomeServer
-from synapse.types import UserID
+from synapse.types import UserID, create_requester
 from synapse.util import Clock
 from twisted.internet.testing import MemoryReactor
 from typing_extensions import override
 
 import tests.unittest as synapsetest
 from synapse_invite_checker import InviteChecker
+from tests.server import FakeChannel
 from tests.test_utils import (
     DOMAIN_IN_LIST,
     DOMAIN2_IN_LIST,
     INSURANCE_DOMAIN_IN_LIST,
     SERVER_NAME_FROM_LIST,
 )
+from tests.test_utils.fake_room_creation import FakeRoom
 
 
 # ruff: noqa: E501
@@ -427,6 +432,9 @@ class FederatingModuleApiTestCase(synapsetest.FederatingHomeserverTestCase):
         self.auth_handler = homeserver.get_auth_handler()
         self.inv_checker: InviteChecker = self.hs.mockmod
 
+        # Map room_id to FakeRoom
+        self.remote_rooms: dict[str, FakeRoom] = {}
+
     @override
     def make_homeserver(self, reactor: MemoryReactor, clock: Clock) -> HomeServer:
         # Mock out the calls over federation.
@@ -492,7 +500,8 @@ class FederatingModuleApiTestCase(synapsetest.FederatingHomeserverTestCase):
             self.inv_checker.permissions_handler.update_permissions(owning_user, perms)
         )
 
-    def _make_join(self, user_id: str, room_id: str) -> dict[str, Any]:
+    def _make_join(self, user_id: str, room_id: str) -> FakeChannel:
+        """Generate the make_join template"""
         users_domain = UserID.from_string(user_id).domain
         channel = self.make_signed_federation_request(
             "GET",
@@ -500,13 +509,35 @@ class FederatingModuleApiTestCase(synapsetest.FederatingHomeserverTestCase):
             f"?ver={self.ROOM_VERSION}",
             from_server=users_domain,
         )
-        self.assertEqual(channel.code, HTTPStatus.OK, channel.json_body)
-        return channel.json_body
+        return channel
 
-    def send_join(self, joining_user: str, room_id: str) -> None:
-        join_result = self._make_join(joining_user, room_id)
+    def send_join(
+        self,
+        joining_user: str,
+        room_id: str,
+        make_join_expected_code: int = HTTPStatus.OK,
+        send_join_expected_code: int = HTTPStatus.OK,
+    ) -> None:
+        """
+        Join a remote user to a local server. Should be a complete make_join/send_join
+        handshake
+        """
+        # First create the make_join that will need to be signed by the 'remote server'
+        join_result_channel = self._make_join(joining_user, room_id)
+        self.assertEqual(
+            join_result_channel.code,
+            make_join_expected_code,
+            f"make_join: {join_result_channel.json_body}",
+        )
 
-        join_event_dict = join_result["event"]
+        # The make_join can fail if a user was not invited to a room, or was otherwise
+        # denied. Since there is nothing to give to the rest of the handshake below, may
+        # as well just return(since the function doesn't return anything anyway
+        if join_result_channel.code != HTTPStatus.OK:
+            return
+
+        # Pull out the template, sign it then give it to our server to process
+        join_event_dict = join_result_channel.json_body["event"]
         joining_users_domain = UserID.from_string(joining_user).domain
         self.add_hashes_and_signatures_from_other_server(
             join_event_dict,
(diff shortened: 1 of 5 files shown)


tests/base.py
```diff
@@ -492,20 +498,32 @@ def add_permission_to_a_user(self, user_to_permit: str, owning_user: str) -> Non
...
+        room_id: str,
+        make_join_expected_code: int = HTTPStatus.OK,
+        send_join_expected_code: int = HTTPStatus.OK,
+    ) -> None:
+        join_result = self._make_join(
+            joining_user, room_id, expected_code=make_join_expected_code
+        )
+        if make_join_expected_code != HTTPStatus.OK:
+            # May as well since it was expected to fail
+            return
```
nico-famedly, line 526: You should assert, tht the expected status code == the actual status code, but I guess the make_join already does that?

tests/test_local_joins.py
```diff
@@ -0,0 +1,131 @@
...
+            self.helper.invite(
+                room_id, self.user_a, self.user_d, tok=self.access_token_a
+            )
+
+        self.helper.join(room_id, self.user_b, tok=self.access_token_b)
+        self.helper.join(room_id, self.user_c, tok=self.access_token_c)
+        self.helper.join(room_id, self.user_d, tok=self.access_token_d)
+
+    @parameterized.expand([("with_invites", True), ("no_invites", False)])
+    def test_joining_private(self, label: str, use_invites: bool) -> None:
```
nico-famedly, line 101: You could probably just split this test into 2, it doesn't really look simpler when you test different things anyway :D

Also re the users even different? Probably 1 or 2 joins are enough.

tests/test_remote_joins.py
```diff
@@ -0,0 +1,507 @@
...
+    FederatingModuleApiTestCase,
+    construct_extra_content,
+)
+from tests.test_utils import DOMAIN_IN_LIST, INSURANCE_DOMAIN_IN_LIST
+
+
+logger = logging.getLogger(__name__)
+
+
+class IncomingRemoteJoinTestCase(FederatingModuleApiTestCase):
```
nico-famedly, line 38: The commit message says, imo, that none of these should have any changes in behaviour, since that is not implemented. What are those tests for?

nico-famedly (approved):
I think this is okay, but I have a few nitpicks and that is a lot of complexity just for tests...

tests/base.py
jason-famedly: Yes, that's right

tests/test_remote_joins.py
```diff
@@ -0,0 +1,507 @@
...
+        creating_user: str,
+        invitee_list: list[str],
+        is_public: bool,
+    ) -> str | None:
+        """
+        Helper to send an api request with a full set of required additional room state
+        to the room creation matrix endpoint.
+        """
+        # Hide the assertion from create_room_as() when the error code is unexpected. It
+        # makes errors for the tests less clear when all we get is the http response
```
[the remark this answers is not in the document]
jason-famedly: This function was a copy paste from other test cases. In a purist sense, I suppose suppressing an assertion that won't raise since there should be nothing blocking creation of an empty room for these tests would be rather redundant. If you want me to remove that I can, but it should never be hit(unless something much worse has happened to the homeserver during setup :D I which case this is small potatoes compared to that). Eventually this will be refactored into the homeserver TestCases themselves

Logically, at least for the test cases this was borrowed from, we didn't care *why* the room failed creation just that it did. The responses that were returned when it was failing were seriously unhelpful. Since that was upstream code in basis, I left it be

tests/test_remote_joins.py
jason-famedly: 1. We didn't actually test against the broader case of a room joins after room creation, which now that we are looking at public/private was probably an oversight. We did test for invites *during* room creation, and how that should prohibit creation of the room
2. Skipping the test when it's known not to pass because of unimplemented functionality instead of making the test pass just to model existing behavior was acceptable behavior? The skips are removed in later PR's

tests/test_remote_joins.py
```diff
@@ -0,0 +1,507 @@
...
+        """
+        Test that the local server can successfully block joining a remote room when
+        there are no invites
+        """
+        remote_room_id = self.create_remote_room(self.remote_pro_user, "10", is_public)
+        assert remote_room_id is not None
+
+        # Public rooms should fail.
+        # Private rooms should also fail because no invite. Should be a 403
+        self.assertRaises(
```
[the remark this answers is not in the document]
jason-famedly: Possibly, but it would require a lot more code complexity I think.

Usually when we check for a specific response code, it's because we are testing our behavior of a local endpoint. We don't do that here. We are expecting either a function to complete correctly(and return an event ID and a stream ordering position, neither of which are useful) or a failure that is a `SynapseError` and don't have access to the complete 'fake channel' we usually get in such other scenarios.

it might be possible to trap the `SynapseError` inside `do_remote_join()`. What would you recommend/would like to see?

tests/test_local_joins.py
jason-famedly: I can split it if you would like. The users should be the same. This is a test only against a PRO server, as the EPA shouldn't allow this functionality, IIRC

tests/base.py
nico-famedly: Then what is the if condition for? Can you either assert there as well or get rid of it?

tests/test_local_joins.py
nico-famedly: I think it would be simpler to read. Less context to have around.

tests/test_remote_joins.py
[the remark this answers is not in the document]
nico-famedly: I think we should not silently swallow assertions, it becomes hard to debug imo.

tests/test_remote_joins.py
nico-famedly: Ah, I didn't notice the skips when I was up here yet. I guess that is fine...

tests/test_remote_joins.py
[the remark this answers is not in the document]
nico-famedly: Considering that do_remote_join is our helper function, I would expect that to have an expected error code parameter and then pass that in and have it assert internally?

tests/test_remote_joins.py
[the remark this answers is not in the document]
jason-famedly: I will adapt them for the current series of tests. If you want that fixed in other tests globally, it will need to be separate work, I think. I can try and make it part of the next major test refactor

tests/test_local_joins.py
jason-famedly: I will split the tests up, but I would like to keep all the users. On the one hand, we are verifying that publicly visible users are not restricted locally. On the other, we are verifying that multiple users can join a local room(maybe I should write such a test case for EPA as well, it seems to be missing. Just invites are checked, not joins)

tests/base.py
jason-famedly: I'm having trouble figuring out what to do here. make_join/send_join is a handshake, so if the 'make' part is expected to fail, then why should I test the 'send' part?

That being said, I pulled the assertion out of `_make_join()` and put it into `send_join()`, but there I still don't know what to do about not exiting early. Can't send the join if it doesn't exist

jason-famedly: See [c8656a7](https://github.com/famedly/synapse-invite-checker/pull/58/commits/c8656a722220c084a506171b7c0608c3ae9cd8b3)

tests/test_local_joins.py
jason-famedly: in [56acf6d](https://github.com/famedly/synapse-invite-checker/pull/58/commits/56acf6df30e1b1a8e86ff75458d766d5c9db281a)

tests/test_remote_joins.py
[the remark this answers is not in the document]
jason-famedly: partly in [56acf6d](https://github.com/famedly/synapse-invite-checker/pull/58/commits/56acf6df30e1b1a8e86ff75458d766d5c9db281a)
and the rest in [0b46548](https://github.com/famedly/synapse-invite-checker/pull/58/commits/0b46548e322c9702a4c7da46fb4b363a6cfcc2cf)

tests/test_remote_joins.py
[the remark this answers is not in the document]
jason-famedly: gave it a try in [c26854d](https://github.com/famedly/synapse-invite-checker/pull/58/commits/c26854d5d28bc0d9dca63bc6192c677852684ce0)

nico-famedly (approved):

tests/base.py
nico-famedly: An assert should exit early, doesn't it? So that return is probably redundant, but whatever.

jason-famedly: Only when it fails. Since are testing that it *does* fail, it won't throw the assert and exit unfortunately

nico-famedly: Ah, right, I got some wires crossed here :D

PR #59, opened 2025-03-06
chore: Forbid local user from inviting remote user to a public room

jason-famedly:
requires #61 
bullet point 2 of famedly/product-management#2925 

Public rooms do not require an invite to join, but placing such an invite can circumvent the required blocking. By not allowing such an invite to be created, this should prevent that circumvention

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -766,6 +766,25 @@ class InviteChecker:
             )
             return errors.Codes.FORBIDDEN
 
+        # Find out if this is a public room
+        # The domains are different, or the first section would have caught it. The same
+        # context as above applies, there may not yet be a room_id if this is a room
+        # creation in progress
+        if room_id:
+            state_mapping: dict[tuple[str, str], EventBase] = (
+                await self.api._storage_controllers.state.get_current_state(
+                    room_id,
+                    StateFilter.from_types([(EventTypes.JoinRules, None)]),
+                )
+            )
+            if event := state_mapping.get((EventTypes.JoinRules, "")):
+                if event.content["join_rule"] == JoinRules.PUBLIC:
+                    logger.debug(
+                        "Forbidding invite to a local public room to a remote user (%s)",
+                        remote_user_id,
+                    )
+                    return errors.Codes.FORBIDDEN
+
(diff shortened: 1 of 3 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -766,6 +766,25 @@ async def user_may_invite(
...
+        # context as above applies, there may not yet be a room_id if this is a room
+        # creation in progress
+        if room_id:
+            state_mapping: dict[tuple[str, str], EventBase] = (
+                await self.api._storage_controllers.state.get_current_state(
+                    room_id,
+                    StateFilter.from_types([(EventTypes.JoinRules, None)]),
+                )
+            )
+            for event in state_mapping.values():
```
nico-famedly, line 780: Shouldn't you be able to access that directly by just indexing using `(EventTypes.JoinRules, None)`?

nico-famedly (commented):
Do we have no tests, that test inviting into a public room after creation? Could you add a short one for that?

jason-famedly:
um, what happened?

jason-famedly:
not sure why this closed. I didn't hit that button

jason-famedly:
open

synapse_invite_checker/invite_checker.py
jason-famedly: Probably. Unless by some bizarre set of circumstances there is not a `join_rules` event.

nico-famedly: dict.get would return None in that case.

jason-famedly:
> Do we have no tests, that test inviting into a public room after creation? Could you add a short one for that?

[c0236b3](https://github.com/famedly/synapse-invite-checker/pull/59/commits/c0236b3eba907f87e2311b1ba0105de8c196888f)
There were, just got mixed up somewhere. Its a partial unblock of one of the newer ones

synapse_invite_checker/invite_checker.py
jason-famedly: Any empty `set_like` iterable would do the same :laughing: Since it would be the only thing in there because of the filter. But sure, let's do a walrus operate so it's a one line adjustment

nico-famedly (approved):

PR #60, opened 2025-03-10
Refactor and test cleanup

jason-famedly:
Some tests are showing their age. Comments were out of date or nor longer relevant, some remote users were missing. Also do some prep work to prepare for removing the Contacts API for EPA mode.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -39,8 +39,10 @@ from twisted.internet.testing import MemoryReactor
 from typing_extensions import override
 
 import tests.unittest as synapsetest
+from synapse_invite_checker import InviteChecker
 from tests.test_utils import (
     DOMAIN_IN_LIST,
+    DOMAIN2_IN_LIST,
     INSURANCE_DOMAIN_IN_LIST,
     SERVER_NAME_FROM_LIST,
 )
@@ -143,8 +145,8 @@ def return_localization(mxid: str) -> str:
         f"@mxid:{DOMAIN_IN_LIST}",
         f"matrix:u/matrixuri%3A{DOMAIN_IN_LIST}",
         f"matrix:u/matrixuri2:{DOMAIN_IN_LIST}",
-        f"matrix:user/gematikuri%3A{DOMAIN_IN_LIST}",
-        f"matrix:user/gematikuri2:{DOMAIN_IN_LIST}",
+        f"matrix:user/gematikuri%3A{DOMAIN2_IN_LIST}",
+        f"matrix:user/gematikuri2:{DOMAIN2_IN_LIST}",
         f"matrix:u/a%3A{SERVER_NAME_FROM_LIST}",
     }:
         return "pract"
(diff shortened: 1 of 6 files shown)


tests/base.py
```diff
@@ -39,8 +39,10 @@
 from typing_extensions import override
 
 import tests.unittest as synapsetest
+from synapse_invite_checker import InviteChecker
 from tests.test_utils import (
     DOMAIN_IN_LIST,
+    DOMAIN_IN_LIST_2,
```
nico-famedly, line 45: Nit, but shouldn't it be DOMAIN2_IN_LIST? Or do we have a second list?

jason-famedly: no second list, no. I can rename that

tests/base.py
```diff
@@ -470,6 +483,15 @@ def add_a_contact_to_user_by_token(
         )
         assert channel.code == 200, channel.result
 
+    def add_permission_to_a_user(self, user_to_permit: str, owning_user: str) -> None:
```
nico-famedly, line 486: Nitpick for a future refactor, why do we duplicate all of this functionality? That seems unnecessary?

jason-famedly: We have two separate major test case homeservers. One for federating models and a lighter one that doesn't. We could *probably* just have the one, if you'd prefer?

tests/test_remote_invites.py
```diff
@@ -61,214 +63,191 @@ async def may_invite(self, inviter: str, invitee: str, roomid: str):
...
-        for inviter in {
-            f"@mxid404:{DOMAIN_IN_LIST}",
+    def test_invite_from_unlisted_user(self) -> None:
+        """
+        Tests that an invite from a remote user that is on the fed list but is otherwise
+        unlisted behaves as expected
+        """
+        for remote_user_id in [
+            f"@example:{DOMAIN_IN_LIST}",
+            "@example:messenger.spilikin.dev",
```
nico-famedly, line 73: Why is this domain hardcoded?

tests/base.py
jason-famedly: Actually, now that you say that: I guess we don't actually have to have permissions available on the non-federated model. Permissions aren't a thing on local servers, right?

tests/test_remote_invites.py
```diff
@@ -61,214 +63,191 @@ async def may_invite(self, inviter: str, invitee: str, roomid: str):
...
+                    == NOT_SPAM
+                ), f"'{remote_user_id}' should be ALLOWED to invite {self.pro_user_a}"
+
+    def test_invite_from_remote_outside_of_fed_list(self) -> None:
+        """Tests that an invite from a remote server not in the federation list gets denied"""
+        for remote_user_id in [
+            f"@example:not-{DOMAIN_IN_LIST}",
+            f"@example2:not-{DOMAIN_IN_LIST}",
+            "@madeupuser:thecornerstore.de",
+            "@unknown:not.in.fed",
```
nico-famedly, line 121: Same here, why are these domains hardcoded?

tests/test_remote_invites.py
jason-famedly: Historically, it just wasn't. I could make that be a constant like the rest in the follow up

tests/test_remote_invites.py
jason-famedly: Those are fake :laughing:

tests/base.py
jason-famedly: Quick test run is showing that ripping it out is erroring because some of the remote tests are using the un-federated model TestCase. Would you rather I go ahead and toss it and move that or wait for the next house cleaning?

nico-famedly (approved):
In general looks fine, but you will want to squash this down and probably word the commit message for that nicely.

tests/base.py
nico-famedly: Wait for the next one probably.

nico-famedly (approved):

PR #61, opened 2025-03-11
Forbid local users being invited to remote public rooms

jason-famedly:
Invites from remote servers usually have some stripped state events added to the request. Parse that data to see if the room being invited to is public, and deny if so

Requires: #58


nico-famedly (approved):
Looks good to me, but the missing None check does trip me up a bit :D

nico-famedly (approved):

PR #63, opened 2025-03-12
chore: Forbid remote users from joining local public rooms

jason-famedly:
This is an experiment, and while it appears to work it does need more extensive auditing of the other code paths used by this third party API callback. I chose to use the one that was associated with `on_make_join_request()` which should be early enough in the join process cycle to avoid leaking information and forbid joining rooms

requires #59


jason-famedly:
`check_event_allowed` auditing(in context of `EventCreationHandler.create_new_client_event()` as the other two uses are knocking related and therefore irrelevant):
handlers/federation.py(6 results):
* [ ] on_make_join_request(): line 915 <- this one is where we use it
* [ ] on_make_leave_request(): line 1188
* [ ] on_make_knock_request(): line 1234 (this one is dumb. it calls both `create_new_client_event()` *and* then `check_event_allowed()` right after)
* [ ] exchange_third_party_invite(): line 1436
* [ ] on_exchange_third_party_invite_request(): line 1520
* [ ] add_display_name_to_third_party_invite(): line 1586

handlers/message.py(1(real) result, other is definition):
* create_event(): line 565
   * from message.py:
      * [ ] _create_and_send_nonmember_event_locked(): line 1041
      * [ ] _send_dummy_event_for_room(): line 2026
   * from room.py
      * [ ] RoomCreationHandler.upgrade_room(): line 185
      * [ ] in a subroutine as part of sending events into a cloned room, these should be ok to be ignored(check if an old room existed that would have been denied now is a thing?)
    * from room_member.py
      * [ ] RoomMemberHandler._local_membership_update(): line 385  <- this one looks interesting
      * [ ] RoomMemberMasterHandler._generate_local_out_of_band_leave(): line 2033

PR #64, opened 2025-03-14
feat: Forbid public room federation by forcing `m.federate` to False

jason-famedly:
Always disable federation with local public rooms by setting `m.federate` to `False` when creating a room. This will override what is set by any client requests.

Provides for a configuration setting to disable.
`override_public_room_federation`: "false" will stop overriding the event, default is "true"

Resolves: famedly/product-management#2926
Resolves: famedly/product-management#2948

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ modules:
         inactive_room_scan:
           enabled: true or false # optional switch to disable the room scan for inactive rooms, defaults to true
           grace_period: see 'Duration Parsing' below # Length of time a room is allowed to have no message activity before it is eligible for deletion. Ignored if 'enabled' is false. Defaults to "26w" which is 6 months
+        override_public_room_federation: true or false, # Forces the `m.federate` flag to be set to False when creating a public room to prevent it from federating. Default is "true", disable with "false"
 ```
 ### Duration Parsing
 Settings labeled as 'duration_parsing' allow for a string representation of the value
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -369,6 +369,14 @@ class InviteChecker:
         _config.inactive_room_scan_options.grace_period_ms = (
             inactive_room_scan_grace_period
         )
+
+        _override_public_room_federation = config.get(
+            "override_public_room_federation", _config.override_public_room_federation
+        )
+        if not isinstance(_override_public_room_federation, bool):
+            msg = "`override_public_room_federation` must be a boolean"
+            raise ConfigError(msg)
+        _config.override_public_room_federation = _override_public_room_federation
         return _config
 
     def after_startup(self) -> None:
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -48,6 +48,7 @@ modules:
         inactive_room_scan:
           enabled: true or false # optional switch to disable the room scan for inactive rooms, defaults to true
           grace_period: see 'Duration Parsing' below # Length of time a room is allowed to have no message activity before it is eligible for deletion. Ignored if 'enabled' is false. Defaults to "26w" which is 6 months
+        override_public_room_federation: true or false, # Allows overriding the `m.federate` setting when creating a public room to prevent federation. Default is "true", disable with "false"
```
nico-famedly, line 51:
```suggestion
        override_public_room_federation: true or false, # Forces the `m.federate` flag to be set to false when creating a public room to prevent it from federating. Default is "true", disable with "false"
```

synapse_invite_checker/invite_checker.py
```diff
@@ -667,17 +675,32 @@ async def on_create_room(
                 errors.Codes.FORBIDDEN,
             )
 
+        # preset can be any of "private_chat", "trusted_private_chat" or "public_chat"
+        # Do not allow "public_chat". Default is based on setting of visibility
+        room_preset: str = request_content.get("preset")
+        # visibility can be either "public" or "private". If not included, it defaults to "private"
+        room_visibility: str = request_content.get("visibility", "private")
+        # Determine based on above that the room is probably public
+        is_public = (
```
nico-famedly, line 684: what if I put join rules in the initial state?

nico-famedly: That is not what I am asking. I was asking what happens if you put a join rule into the initial_state list of the /createRoom call.

nico-famedly: Ah, the code was folded and I didn't notice it, seems fine! :)

synapse_invite_checker/invite_checker.py
```diff
@@ -698,11 +699,30 @@ async def on_create_room(
...
+            can_federate = creation_content.setdefault("m.federate", True)
+
+            if can_federate and is_public:
+                if self.config.override_public_room_federation:
+                    logger.debug("Overriding `m.room.create` to disable federation")
+                    request_content["creation_content"]["m.federate"] = False
+                else:
+                    logger.warning(
+                        "Room creation with a public room allowed to federate detected."
+                    )
```
nico-famedly, line 716: I think this modifies the request_content in any case. Could we somehow skip modifying the request_content, when we don't need to?

jason-famedly: See what you think about [7839b39](https://github.com/famedly/synapse-invite-checker/pull/64/commits/7839b39ec9bc8bdc7907b809238bdea4430978f0)

tests/test_remote_joins.py
```diff
@@ -267,17 +255,113 @@ def test_local_room_remote_pro_with_invites_no_permission(
...
+        super().prepare(reactor, clock, homeserver)
+        #  "a" is a practitioner
+        #  "b" is an organization
+        #  "c" is an 'orgPract'
+        self.user_a = self.register_user("a", "password")
+        self.access_token_a = self.login("a", "password")
+        # self.user_b = self.register_user("b", "password")
+        # self.access_token_b = self.login("b", "password")
+        # self.user_c = self.register_user("c", "password")
+        # self.access_token_c = self.login("c", "password")
```
nico-famedly, line 293: dead code

nico-famedly (approved):

tests/test_remote_joins.py
jason-famedly: I'll add it to the wish list

PR #65, opened 2025-03-17
chore: Only kick insured users from their rooms, do not purge

jason-famedly:
Resolves: famedly/product-management#2947

Instead of just outright purging an EPA server room, 'block' then 'leave' the room with local users. By blocking the room, no further activity should be possible but will allow leaves to still take place

Apparently, server notices rooms were being included in the room scan before. They are now excluded from the EPA room scan but will still be included for inactive purging

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -52,7 +52,16 @@ from synapse.http.server import JsonResource
 from synapse.module_api import NOT_SPAM, ModuleApi, errors
 from synapse.server import HomeServer
 from synapse.storage.database import LoggingTransaction, make_conn
-from synapse.types import Requester, RoomID, ScheduledTask, StateMap, TaskStatus, UserID
+from synapse.types import (
+    Requester,
+    RoomID,
+    ScheduledTask,
+    StateMap,
+    TaskStatus,
+    UserID,
+    create_requester,
+)
+from synapse.storage.roommember import RoomsForUser
 from synapse.types.handlers import ShutdownRoomParams
 from synapse.types.state import StateFilter
 from synapse.util.metrics import measure_func
@@ -801,6 +810,57 @@ class InviteChecker:
             statuses=statuses,
         )
 
+    async def kick_all_local_users(self, room_id: str) -> None:
+        """
+        First block the room from further interaction from this server, then kick all
+        local members
+
+        Args:
+            room_id:
+        """
+        requester_user_id = (
+            f"@_synapse-invite-checker:{self.api._hs.config.server.server_name}"
+        )
+
+        # Block the room, this prevents inadvertent rejoins and stray events/messages
+        # from becoming part of the room after the fact but will still allow the user
+        # leaving to occur.
+        # NOTE: This call is idempotent, so even if some unforeseen error occurs that
+        # causes the user to fail to leave the room, it will still be blocked and
+        # calling this again won't unblock it or otherwise interfere.
+        await self.api._store.block_room(room_id, requester_user_id)
+
+        users = await self.api._store.get_local_users_related_to_room(room_id)
+        for user_id, membership in users:
+            # If the user is not in the room (or is banned), nothing to do.
+            if membership not in (Membership.JOIN, Membership.INVITE, Membership.KNOCK):
+                continue
+
+            logger.info("Kicking %s from %s...", user_id, room_id)
+            # Use the actual user as a puppet. Don't use the auto-kicker
+            # user id above, as it won't pass room auth
+            target_requester = create_requester(user_id, authenticated_entity=user_id)
+
+            try:
+                # Kick users from room
+                (
+                    _,
+                    stream_id,
+                ) = await self.api._hs.get_room_member_handler().update_membership(
+                    requester=target_requester,
+                    target=target_requester.user,
+                    room_id=room_id,
+                    action=Membership.LEAVE,
+                    content={},
+                    ratelimit=False,
+                    require_consent=False,
+                )
+
+            except Exception:
+                logger.exception(
+                    "Failed to kick user '%s' from room %s", user_id, room_id
+                )
+
     async def schedule_room_for_purge(self, room_id: str) -> None:
         """
         Schedules the deletion of a room from Synapse's database after kicking all users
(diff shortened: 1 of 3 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -801,6 +810,55 @@ async def get_delete_tasks_by_room(self, room_id: str) -> list[ScheduledTask]:
...
+        # calling this again won't unblock it or otherwise interfere.
+        await self.api._store.block_room(room_id, requester_user_id)
+
+        users = await self.api._store.get_local_users_related_to_room(room_id)
+        for user_id, membership in users:
+            # If the user is not in the room (or is banned), nothing to do.
+            if membership not in (Membership.JOIN, Membership.INVITE, Membership.KNOCK):
+                continue
+
+            logger.info("Kicking %r from %r...", user_id, room_id)
```
nico-famedly, line 837: why are you mixing s and r?

jason-famedly: Because for strings it doesn't really matter. I can change it tho, at first I thought they were the `UserID` and `RoomID` classes :laughing:

jason-famedly: Fixed up now, all are `%s`

nico-famedly: (it leaves people scratching their heads)

synapse_invite_checker/invite_checker.py
```diff
@@ -860,16 +936,9 @@ async def room_scan(self) -> None:
                         + self.config.insured_room_scan_options.grace_period_ms
                         <= self.api._hs.get_clock().time_msec()
                     ):
-                        rooms_to_purge.add(room_id)
-
-        # may as well get these moving
-        for room_id in rooms_to_purge:
-            await self.schedule_room_for_purge(room_id)
+                        await self.kick_all_local_users(room_id)
```
nico-famedly, line 939: Couldn't you have just used the existing purge room call, but instead of passing "purge: True" passing False? It should still kick everyone then?

jason-famedly: It would still 'forget' the room

nico-famedly: Oh, right, makes sense, thanks!

synapse_invite_checker/invite_checker.py
```diff
@@ -845,11 +903,29 @@ async def room_scan(self) -> None:
...
+            # still check all the inactive rooms to purge below
+            server_notice_rooms = set(room.room_id for room in server_notice_rooms_list)
+
         if self.config.insured_room_scan_options.enabled:
             for room_id in all_room_ids:
-                # only purge rooms that only have EPA hosts in them
-
+                # Server notice rooms are exempt, don't want to necessarily hide that
+                # a notice showed up before the user has a chance to see it.
+                if room_id in server_notice_rooms:
```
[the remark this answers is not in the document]
jason-famedly: I actually considered this. A batch collection of all the server notice rooms is a lot more efficient

https://github.com/famedly/synapse/blob/cbd0174c7acfbb42f83acad82684cb6ef5e8d84a/synapse/storage/databases/main/roommember.py#L450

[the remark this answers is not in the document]
nico-famedly: can you explain in the comment then, that you are explicitly not using that function? Just so that the next person doesn't simplify the logic?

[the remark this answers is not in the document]
jason-famedly: Sure, should have it attached now

synapse_invite_checker/invite_checker.py
```diff
@@ -801,6 +810,55 @@ async def get_delete_tasks_by_room(self, room_id: str) -> list[ScheduledTask]:
             statuses=statuses,
         )
 
+    async def kick_all_local_users(self, room_id: str) -> None:
+        """
+        First block the room from further interaction from this server, then cause all
+        local members to set their membership to 'leave'
```
nico-famedly, line 816:
```suggestion
        First block the room from further interaction from this server, then kick all local users.
```

I don't think you need to explain how kicking works...

synapse_invite_checker/invite_checker.py
```diff
@@ -801,6 +810,55 @@ async def get_delete_tasks_by_room(self, room_id: str) -> list[ScheduledTask]:
...
+
+        Args:
+            room_id:
+        """
+        requester_user_id = f"@auto-kicker:{self.api._hs.config.server.server_name}"
```
nico-famedly, line 821:
```suggestion
        requester_user_id = f"@_synapse-invite-checker:{self.api._hs.config.server.server_name}"
```

Underscore is a reserved namespace afaik and I think it makes sense to use the module name here?

synapse_invite_checker/invite_checker.py
```diff
@@ -801,6 +810,55 @@ async def get_delete_tasks_by_room(self, room_id: str) -> list[ScheduledTask]:
...
+                    room_id=room_id,
+                    action=Membership.LEAVE,
+                    content={},
+                    ratelimit=False,
+                    require_consent=False,
+                )
+
+            except Exception:
+                logger.exception(
+                    "Failed to 'leave' user '%s' from room %s", user_id, room_id
```
nico-famedly, line 859:
```suggestion
                    "Failed to kick user '%s' from room %s", user_id, room_id
```

synapse_invite_checker/invite_checker.py
```diff
@@ -845,11 +903,32 @@ async def room_scan(self) -> None:
...
+                    self.api._hs.config.servernotices.server_notices_mxid,
+                    [Membership.JOIN],
+                    [],
+                )
+            )
+            # The server_notice_rooms_list is more complex than we need, pare it down to
+            # room_id's only. A set() is conveniently O(1) complexity. Also probably not
+            # a good idea to just remove these outright from the complete room list, to
+            # still check all the inactive rooms to purge below
+            server_notice_rooms = set(room.room_id for room in server_notice_rooms_list)
```
nico-famedly, line 923:
```suggestion
            server_notice_rooms = {room.room_id for room in server_notice_rooms_list}
```

Probably makes no difference, but afaik we use set literals elsewhere.

synapse_invite_checker/invite_checker.py
```diff
@@ -845,11 +903,32 @@ async def room_scan(self) -> None:
...
+
         if self.config.insured_room_scan_options.enabled:
             for room_id in all_room_ids:
-                # only purge rooms that only have EPA hosts in them
-
+                # Server notice rooms are exempt, don't want to necessarily hide that
```
nico-famedly, line 927: We should also exempt them from the inactivity purge, I think?

nico-famedly: In which case we could just remove all rooms from all_room_ids that are in server_notice_rooms.

nico-famedly (dismissed):

synapse_invite_checker/invite_checker.py
jason-famedly: I thought I recalled somewhere that this was discussed and decided that server notice rooms were still eligible for cleanup on the basis that they should be purged in the instance of an employee leaving? Something like that, I think

jason-famedly: But if you do want to go down that road, let's do it separate

synapse_invite_checker/invite_checker.py
```diff
@@ -845,11 +903,32 @@ async def room_scan(self) -> None:
...
+                await self.api._store.get_rooms_for_local_user_where_membership_is(
+                    self.api._hs.config.servernotices.server_notices_mxid,
+                    [Membership.JOIN],
+                    [],
+                )
+            )
+            # The server_notice_rooms_list is more complex than we need, pare it down to
+            # room_id's only. A set() is conveniently O(1) complexity. Also probably not
+            # a good idea to just remove these outright from the complete room list, to
+            # still check all the inactive rooms to purge below
```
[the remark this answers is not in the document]
jason-famedly: I guess I was particularly wordy that day, that is much better

synapse_invite_checker/invite_checker.py
jason-famedly: Yeah, sounds good

synapse_invite_checker/invite_checker.py
jason-famedly: It gets written to a database table that is never looked at again, and no processing is done on it :laughing: I changed it to this anyway

synapse_invite_checker/invite_checker.py
jason-famedly: Fair, patched up

synapse_invite_checker/invite_checker.py
nico-famedly: But currently we purge server notice rooms also for employees, that are still there... But I guess, whatever.

synapse_invite_checker/invite_checker.py
nico-famedly: But it would be weird to use a registerable userid, imo :)

nico-famedly: Also makes debugging easier!

PR #66, opened 2025-03-17
feat: apply permissions locally

nico-famedly:
This is required by the new TIM specifications. For this reason we are defaulting to "allow all" as the default permission as otherwise users would not be contactable when they list themselves in the directory. This can then be changed in the client.

Since we now default to "allow all" there is no need to migrate any contacts. As such most of the database code is removed. For the same reason the contacts API is removed.

We may want to allow different default settings on the server side in the future. This will need a lot of test changes though. For now I kept the test changes "minimal", but we can probably simplify a lot now since we only need to test all the invite permission permutations and that insured people can't invite each other.

This also removes the old localization support, where a user would be contactable just from being listed in the VZD.


jason-famedly (approved):
I guess this ok

PR #69, opened 2025-03-18
chore: Forbid changing room state event on EPA server

jason-famedly:
Resolves famedly/product-management#2941

Now it is forbidden to change the room's 
* `m.room.join_rules` to "public" 
* `m.room.history_visibility` to "world_readable"

Folded some previous changes that were looking at `initial_state` during room creation into this so it was no duplicated.

Rolled the prevention of changing `m.room.history_visibility` during room creation into this as well, I think it got skipped before somehow

New configuration setting:
```
prohibit_world_readable_rooms: "true" or "false"  # Defaults to "true", use "false" to disable checking for this
```

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -34,7 +34,9 @@ from OpenSSL.crypto import (
 from synapse.api.constants import (
     AccountDataTypes,
     Direction,
+    EventContentFields,
     EventTypes,
+    HistoryVisibility,
     JoinRules,
     Membership,
     RoomCreationPreset,
@@ -194,6 +196,9 @@ class InviteChecker:
         self.api.register_third_party_rules_callbacks(
             on_upgrade_room=self.on_upgrade_room
         )
+        self.api.register_third_party_rules_callbacks(
+            check_event_allowed=self.check_event_allowed
+        )
 
         dbconfig = None
         for dbconf in api._store.config.database.databases:
@@ -352,6 +357,14 @@ class InviteChecker:
             msg = "`override_public_room_federation` must be a boolean"
             raise ConfigError(msg)
         _config.override_public_room_federation = _override_public_room_federation
+
+        _prohibit_world_readable_rooms = config.get(
+            "prohibit_world_readable_rooms", _config.prohibit_world_readable_rooms
+        )
+        if not isinstance(_prohibit_world_readable_rooms, bool):
+            msg = "`prohibit_world_readable_rooms` must be a boolean"
+            raise ConfigError(msg)
+        _config.prohibit_world_readable_rooms = _prohibit_world_readable_rooms
         return _config
 
     def after_startup(self) -> None:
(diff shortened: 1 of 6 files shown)


nico-famedly:
https://github.com/famedly/product-management/issues/2948 said to also prevent this for pro-mode depending on if the room is allowed to federate. Should I reopen that issue or can you just fold it in here?

jason-famedly:
> [famedly/product-management#2948](https://github.com/famedly/product-management/issues/2948) said to also prevent this for pro-mode depending on if the room is allowed to federate. Should I reopen that issue or can you just fold it in here?

I think I can fold it in, access to the room creation event is handy in this case

synapse_invite_checker/invite_checker.py
```diff
@@ -506,6 +511,70 @@ async def user_may_join_room(
...
+                event.type == EventTypes.JoinRules
+                and event.content["join_rule"] == JoinRules.PUBLIC
+            ):
+                creation_event = context.get((EventTypes.Create, ""))
+                # `m.federate` defaults to True if unspecified. This allows a convenient
+                # method to determine that this is a public room, since private rooms are
+                # the only ones allowed to have `m.federate` set to True and public rooms
+                # can not exist on EPA servers
+                # Accessing "content" here in this way feels really strange, but at least
+                # doesn't error like when I try to access the property directly
```
nico-famedly, line 550: Please explain what the error is. Otherwise this comment just adds confusion.

synapse_invite_checker/invite_checker.py
```diff
@@ -506,6 +511,70 @@ async def user_may_join_room(
...
+            # Prevent changing "m.room.join_rules"
+            if (
+                event.type == EventTypes.JoinRules
+                and event.content["join_rule"] == JoinRules.PUBLIC
+            ):
+                creation_event = context.get((EventTypes.Create, ""))
+                # `m.federate` defaults to True if unspecified. This allows a convenient
+                # method to determine that this is a public room, since private rooms are
+                # the only ones allowed to have `m.federate` set to True and public rooms
+                # can not exist on EPA servers
```
nico-famedly, line 548: I don't think this comment is true? We have a config flag for that after all, so we shouldn't assume this is always true.

synapse_invite_checker/invite_checker.py
```diff
@@ -506,6 +511,70 @@ async def user_may_join_room(
...
+                        None,
+                    )
+
+            # Prevent changing "m.room.history_visibility"
+            elif (
+                event.type == EventTypes.RoomHistoryVisibility
+                and event.content["history_visibility"]
+                == HistoryVisibility.WORLD_READABLE
+            ):
+                old_history_vis = context.get((EventTypes.RoomHistoryVisibility, ""))
```
nico-famedly, line 568: I don't think we need the dependency on the old history visibility.

nico-famedly: Otoh, the history visibility validation should probably have a config flag. It is a bit risky.

synapse_invite_checker/invite_checker.py
jason-famedly: It's forced now for public rooms to have `m.federate` be False for public rooms, and public rooms can't/won't exist on EPA servers. Doesn't matter tho now, we short circuit out upon detection of EPA server regardless of that flag

synapse_invite_checker/invite_checker.py
jason-famedly: Left over notes to myself from my typechecking being weird. Something about the 'content' property being None when it can't really be and `__get_item__()` not be declared. Not important now so it's gone

synapse_invite_checker/invite_checker.py
```diff
@@ -506,6 +511,70 @@ async def user_may_join_room(
...
+                # method to determine that this is a public room, since private rooms are
+                # the only ones allowed to have `m.federate` set to True and public rooms
+                # can not exist on EPA servers
+                # Accessing "content" here in this way feels really strange, but at least
+                # doesn't error like when I try to access the property directly
+                federated_flag = creation_event["content"].get(
+                    EventContentFields.FEDERATE, True
+                )
+                old_join_rules = context.get((EventTypes.JoinRules, ""))
+                if federated_flag and old_join_rules:
```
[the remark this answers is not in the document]
jason-famedly: I think I got this right now. Although I'm not entirely getting my head around "keeping the room public"

synapse_invite_checker/invite_checker.py
jason-famedly: config flag added, old history vis event dependency removed

synapse_invite_checker/invite_checker.py
```diff
@@ -506,6 +519,59 @@ async def user_may_join_room(
...
+            )
+            # Remember to account for the override disabler
+            if federated_flag and self.config.override_public_room_federation:
+                return False, None
+
+        # If configured, forbid "m.room.history_visibility" to be set as "world_readable"
+        elif (
+            self.config.prohibit_world_readable_rooms
+            and event.type == EventTypes.RoomHistoryVisibility
+            and event.content["history_visibility"] == HistoryVisibility.WORLD_READABLE
```
nico-famedly, line 569: Does that actually work? The documentation says not to do that, but use event.get_dict() and return the new event. I think this really fits better in on_create_room.

synapse_invite_checker/invite_checker.py
```diff
@@ -506,6 +519,63 @@ async def user_may_join_room(
...
+        if not event.is_state() or not self.api.is_mine(event.sender):
+            # We only touch state events, and never anything from another server
+            return True, None
+        # Important Note: This callback also runs during room creation, and may end up
+        # being appropriate for checking the same things we check in `on_create_room()`.
+        # However, this callback is run for *every single event created* and should be
+        # kept as "light weight" as possible
+        # By checking for the creation event right up front, it also helps to guard
+        # against accidentally triggering the condition on the creation event itself
+        if event.type == EventTypes.Create:
```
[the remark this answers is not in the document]
jason-famedly: Good catch, updated to remove that

synapse_invite_checker/invite_checker.py
jason-famedly: does...what actually work? OH! No, we are not modifying the event, we just deny it. I'm not going to have the chance of a borked up room being made. This way it combines both forbidding the room from being created but also forbidding changing it after-the-fact by trying to change state

jason-famedly: If I comment out that whole section, the relevant tests fail as the event/room creation are allowed:
```
FAILED tests/test_createrooms_local.py::LocalProModeCreateRoomTest::test_create_room_then_modify_history_visibility_0_private - AssertionError: Expected: 403, got: 200, resp: b'{"event_id":"$ToNJGMzYE0Y-V-lPk8prqsU4mXSEsI5iLWLyRC-SK-A"}'
FAILED tests/test_createrooms_local.py::LocalProModeCreateRoomTest::test_create_room_then_modify_history_visibility_1_public - AssertionError: Expected: 403, got: 200, resp: b'{"event_id":"$v7VfphPgbrmq0YzhaHtxXJYjBcVio87wKzf8EFyzvLU"}'
FAILED tests/test_createrooms_local.py::LocalEpaModeCreateRoomTest::test_create_room_then_modify_history_visibility - AssertionError: Expected: 403, got: 200, resp: b'{"event_id":"$YD_laqzC0qWJKMtPFvCfwrOV6R6S2tN9G7CpwWGKbYA"}'
FAILED tests/test_createrooms_local.py::LocalEpaModeCreateRoomTest::test_create_room_with_modified_history_visibility - AssertionError: Private room should NOT have been created assert '!NWMeewLSHiwsBkoEfc:ti-messengertest.dev.ccs.gematik.solutions' is None
```

nico-famedly: Ah, yeah, I misread that, because I was doing other stuff on the side, my bad! :D

nico-famedly (approved):

PR #70, opened 2025-03-25
feat: Allow for setting default permissions in configuration

jason-famedly:
New configuration settings allow for establishing the default permissions
```
defaultPermissions:
  defaultSetting: "allow all" or "block all" # "allow all" is set as default
  userExceptions:
    "@example-user:server.com": # this is an example, NOTE the double quotes are necessary because of the '@' sign and the ':' at the end of the entry
  serverExceptions:
    "@LOCAL_SERVER@": # A useful(but not necessary) template variable to designate the local server as an exception
    example.com: # an example, the ':' are necessary just as in 'userExceptions'
  groupExceptions:
  - groupName: isInsuredPerson # The only possible option in groupExceptions at this time. NOTE the '-' at the beginning of the entry to designate this is a list
```
Most useful examples are probably going to be:
* allow all but deny insurance servers
```
defaultPermissions:
  defaultSetting: allow all
  groupExceptions:
  - groupName: isInsuredPerson
```
* block all but local server:
```
defaultPermissions:
  defaultSetting: block all
  serverExceptions:
    "@LOCAL_SERVER@":
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,16 @@ modules:
         federation_list_client_cert: "tests/certs/client.pem", # path to a pem encoded client certificate for mtls, required if federation list url is https
         gematik_ca_baseurl: "https://download-ref.tsl.ti-dienste.de/", # the baseurl to the ca to use for the federation list, required
         tim-type: "epa" or "pro", # Patient/Insurance or Professional mode, defaults to "pro" mode. Optional currently, but will be required in a later release
+        defaultPermissions: # see 'defaultPermissions' below. The server defaults for new users or existing users with no permissions already set. Other than the noted default for 'defaultSetting', no other defaults are established
+          defaultSetting: "allow all" or "block all" # Default "allow all"
+          serverExceptions:
+            "<server_name>": # The server names to include. Note the ':' on the end and that double quotes are needed around server names
+            "@LOCAL_SERVER@": # A special option to template the local server into without having to know its name. Note that the double quotes are required for this special case.
+          userExceptions:
+            "<mxid>": # Any users that should be an exception to the 'defaultSetting'.
+            "@user:some_server.com": # An example. Note the ':' on the end and that double quotes are needed around user names
+          groupException:
+          - groupName: "isInsuredPerson" # For the moment, the only option. Note the double quotes and the hyphen at the start of the line
         allowed_room_versions: # The list(as strings) of allowed room versions. Currently optional, defaults are listed
           - "9"
           - "10"
@@ -50,6 +60,30 @@ modules:
         override_public_room_federation: true or false, # Forces the `m.federate` flag to be set to False when creating a public room to prevent it from federating. Default is "true", disable with "false"
         prohibit_world_readable_rooms: true or false, # Prevent setting any rooms history visibility as 'world_readable'. Defaults to "true"
         block_invites_into_dms: true or false, # Prevent invites into existing DM chats. Defaults to true
+```
+### defaultPermissions
+
+For establishing the default permissions for the users on this server. As the simplest
+example:
+```yaml
+defaultPermissions:
+  defaultSetting: "allow all"
+```
+This is what the default will be if no setting is entered for this section.
+
+an example to allow all communication except for insured users
+```yaml
+defaultPermissions:
+  defaultSetting: "allow all"
+  groupException:
+    - groupName: "isInsuredPerson"
+```
+and an example of blocking all communication except for users on the local server
+```yaml
+defaultPermissions:
+  defaultSetting: "block all"
+  serverExceptions:
+    "@LOCAL_SERVER@":
 ```
 
 ### Duration Parsing
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -15,7 +15,7 @@
 import base64
 import functools
 import logging
-from collections.abc import Callable
+from collections.abc import Callable, Collection
 from contextlib import suppress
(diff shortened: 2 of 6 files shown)


jason-famedly:
oops, still need to add those examples to the README. TODO

README.md
```diff
@@ -37,6 +37,16 @@ modules:
         federation_list_client_cert: "tests/certs/client.pem", # path to a pem encoded client certificate for mtls, required if federation list url is https
         gematik_ca_baseurl: "https://download-ref.tsl.ti-dienste.de/", # the baseurl to the ca to use for the federation list, required
         tim-type: "epa" or "pro", # Patient/Insurance or Professional mode, defaults to "pro" mode. Optional currently, but will be required in a later release
+        defaultPermissions: # The server defaults for new users or existing users with no permissions already set. Other than the noted default for 'defaultSetting', no other defaults are established
+          defaultSetting: "allow all" or "block all" # Default "allow all"
+          serverExceptions:
+            <server_name>: # The server names to include. Note the ':' on the end
```
nico-famedly, line 43: I think it makes sense to just quote all of the strings here for consistency :D

synapse_invite_checker/invite_checker.py
```diff
@@ -663,6 +684,21 @@ async def on_create_room(
...
+        device_id: str | None,
+        initial_display_name: str | None,
+        request_info: Collection[tuple[str | None, str]],
+        auth_provider_id: str | None = None,
+    ) -> Literal["NOT_SPAM"] | errors.Codes:
+        # Setting initial permissions is as simple as calling the get_permissions
+        # function. It will be populated based on the related configuration settings.
+        current_perms = await self.permissions_handler.get_permissions(user_id)
+        # Then just save them back. Done deal.
+        await self.permissions_handler.update_permissions(user_id, current_perms)
```
nico-famedly, line 699: I don't think we should write to the database, if the permissions are already set.

synapse_invite_checker/permissions.py
```diff
@@ -36,20 +36,20 @@ def __init__(
...
     async def get_permissions(self, user_id: str) -> PermissionConfig:
         config_type = await self.get_config_type_from_mxid(user_id)
         account_data = await self.account_data_manager.get_global(
             user_id, config_type.value
         )
         if account_data is None:
-            permissions = PermissionConfig(
-                defaultSetting=PermissionDefaultSetting.ALLOW_ALL,
-            )
+            permissions = self.default_perms
```
nico-famedly, line 52: I think we should write the default permissions to account data here if the required key "defaultSetting" is not present. That way the account data calls are next to each other and we don't write to the db when not necessary.

jason-famedly: I like that idea, good call

README.md
jason-famedly: I don't know. Yaml doesn't really require it, and since that may make it confusing all around for the other sections in the README. It only seemed to be necessary to deal with `@` being an invalid char when it wasn't quoted.

(And for the `defaultSetting` examples, the only reason I quoted them at all was to differentiate to the reader that the string was either block all or allow all and the 'and' was part of neither. I guess I could rewrite that bit to have it in the comment off the end)

If you think it's best, I don't know enough about how the configuration is generated to give a substantial opinion. I do concern myself with how ugly escaping strings can look when trying to read what is done

nico-famedly: I think it is good to quote yaml values, that can be arbitrary strings. That way you don't run into pitfalls that easily. For example I don't know what happens, if my server name includes a port and is specified in the exceptions. Do I need to quote it then? Better quote it always!

nico-famedly: Same with MXIDs btw. They start with an @ too.

synapse_invite_checker/invite_checker.py
```diff
@@ -663,6 +684,20 @@ async def on_create_room(
...
+    async def check_login_for_spam(
+        self,
+        user_id: str,
+        device_id: str | None,
+        initial_display_name: str | None,
+        request_info: Collection[tuple[str | None, str]],
+        auth_provider_id: str | None = None,
+    ) -> Literal["NOT_SPAM"] | errors.Codes:
+        # Setting initial permissions is as simple as calling the get_permissions
+        # function. It will be populated based on the related configuration settings.
```
nico-famedly, line 696:
```suggestion
        # Default permissions are populated automatically when fetching them. This ensures users can see the default permissions in their client when they sign in.
```

synapse_invite_checker/permissions.py
nico-famedly: So you disagree with setting it based on if the defaultSetting is set? Since that is a required field, it should probably be populated everytime. This then catches also cases where someone sends "{}" as their permissions.

nico-famedly (dismissed):

synapse_invite_checker/permissions.py
jason-famedly: Oh, I misread that. Just a minute

jason-famedly: There, how about that change instead? Catch both `None`(so an unset permissions) and `{}`(in case of client mishap) and then also watching for a misformed `defaultSetting`(so either a `None` or an empty `{}` again)

synapse_invite_checker/invite_checker.py
jason-famedly: (almost forgot this, added it too)

nico-famedly (approved):

nico-famedly (approved):

PR #71, opened 2025-03-25
chore: Allow to disable the DM check for invites

jason-famedly:
Resolves: famedly/product-management#2990

Adds configuration option to disable the DM check
```
block_invites_into_dms: true or false # defaults to "true"
```


nico-famedly (dismissed):

jason-famedly:
pushed something here that didn't belong, please re-review

PR #74, opened 2025-03-27
chore: Test cleanups

jason-famedly:


jason-famedly:
Encountered an interesting problem while doing the first part of the wishlist: A public room created by a system administrator forces the `m.federate` flag to True, as the code that forces that to False is bypassed. Now that the join rules are enforced, this causes a room to not be created.

Since at the time the admin bypass was installed in `on_create_room()` public rooms were not a thing, can this be removed now?

cc @nico-famedly

nico-famedly:
No, I think the admin should just be exempt from the other check as well? Dunno.

nico-famedly:
At least we should exempt the admin for the room version check and invite count check still.

tests/base.py
```diff
@@ -595,3 +596,55 @@ def login(
...
+                else:
+                    extra_content.update({key: value})
+
+        # Hide the assertion from create_room_as() when the error code is unexpected. It
+        # makes errors for the tests less clear when all we get is the http response,
+        # because then we are not sure which exact test used is the failure(especially
+        # when creating many rooms). Instead, use a simple binary condition; either we
+        # get a room_id or None. This allows the test itself to let us know which test
+        # failed.
+        with contextlib.suppress(AssertionError):
```
nico-famedly, line 645: This makes it somewhat hard to figure out, what the assertion was. Can we at least log it?

tests/test_utils/fake_room_creation.py
```diff
@@ -31,6 +31,18 @@
 from synapse.types import RoomID
 from synapse.util import Clock, stringutils
 
+"""
+NOTES for robustness:
+    Re-using a room exposed a flaw. The FakeRoom thinks that the last event
```
nico-famedly, line 36: What am I supposed to do about that? Did you intend to just say, that rooms shouldn't be reused or what?

tests/base.py
jason-famedly: During the test? I suppose. The point is that we need to know that a room was made or not, not why it wasn't made. Getting the room ID was sufficient. If we use the expected code assertion built into helper, we don't get to control the assertion or it's message on failure

nico-famedly (commented):
Looks good to me in general, just a few nitpicks.

tests/test_utils/fake_room_creation.py
jason-famedly: I think so. Notes from past me to future me of *why*

tests/base.py
jason-famedly: What would you like the logging to say?

nico-famedly: The assertion message? Then it will be printed only when the test fails and you can still find it instead of having to know, that such assertions are suppressed.

tests/test_utils/fake_room_creation.py
nico-famedly: My problem is mostly, this looks like something I should do something about, but it isn't really actionable. So can you write it in a style, where the reader knows upfront why the note is there? Is this  future todo, is this a design constraint, does it matter in how I use the API?

jason-famedly: I think I'll just give a brief bit in the docstring, this looks wordy and largely irrelevant after having been away from it for a month

tests/base.py
jason-famedly: How's this look? The WARNING is new the other is pre-existing. The problem I see is that these will already exist all over the logging anyway as there are rooms we are expecting to fail. Most of the time, we don't care *why* the room failed, just that it did
```
INFO     synapse.http.server:server.py:130 <SynapseRequest at 0x7545fed974a0 method='POST' uri='/_matrix/client/r0/createRoom?access_token=<redacted>' clientproto='1.1' site='tim.test.gematik.de'> SynapseError: 403 - When creating a room, a maximum of one participant can be invited directly
INFO     synapse.access.http.fake:site.py:508 127.0.0.1 - tim.test.gematik.de - {@a:tim.test.gematik.de} Processed request: 0.009sec/0.000sec (0.004sec, 0.000sec) (0.001sec/0.002sec/4) 110B 403 "POST /_matrix/client/r0/createRoom?access_token=<redacted> 1.1" "-" [0 dbevts]
WARNING  tests.base:base.py:655 create_room_as failed to create room, this may have been expected: AssertionError({'version': b'1.1', 'code': b'403', 'reason': b'Forbidden', 'headers': Headers({b'Server': [b'1'], b'Date': [b'Mon, 28 Apr 2025 12:32:54 GMT'], b'Content-Type': [b'application/json'], b'Cache-Control': [b'no-cache, no-store, must-revalidate'], b'Access-Control-Allow-Origin': [b'*'], b'Access-Control-Allow-Methods': [b'GET, HEAD, POST, PUT, DELETE, OPTIONS'], b'Access-Control-Allow-Headers': [b'X-Requested-With, Content-Type, Authorization, Date'], b'Access-Control-Expose-Headers': [b'Synapse-Trace-Id, Server']}), 'body': b'{"errcode":"M_FORBIDDEN","error":"When creating a room, a maximum of one participant can be invited directly"}', 'done': True})
```

nico-famedly (approved):

PR #75, opened 2025-04-01
chore: Test permissions a little more robustly

jason-famedly:
Part of famedly/product-management#2956

> * Refactor permission test cases to test all variants independent of the default and structured by the current set of requirements instead of ordered by hba user vs normal user, etc. Ideally we would have a table of invitee permissions default, server and user exceptions, inviter and invitee and expected result. (We still need SOME tests for the default case of course, see https://github.com/famedly/product-management/issues/2967 )

I do not think this got completely done as expected, but there are some decent improvements here. I dislike that the line removal from #74 seems to have been for nothing

Depends on
* #74

### DIFF
--- a/tests/test_local_invites.py
+++ b/tests/test_local_invites.py
@@ -12,12 +12,15 @@
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
+from http import HTTPStatus
 from typing import Any
 
+from parameterized import parameterized
 from synapse.server import HomeServer
 from synapse.util import Clock
 from twisted.internet.testing import MemoryReactor
 
+from synapse_invite_checker.types import PermissionConfig, PermissionDefaultSetting
 from tests.base import FederatingModuleApiTestCase
 from tests.test_utils import INSURANCE_DOMAIN_IN_LIST_FOR_LOCAL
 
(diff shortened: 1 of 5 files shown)


tests/test_local_invites.py
```diff
@@ -31,11 +35,257 @@ class LocalProModeInviteTest(FederatingModuleApiTestCase):
...
+        is_public: bool,
+        expected_result: int,
+    ) -> None:
+        room_b = self.create_local_room(
+            self.user_b,
+            [],
+            is_public=is_public,
+        )
+        assert room_b is not None, "Room should have been created"
+        room_c = self.create_local_room(
```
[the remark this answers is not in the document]
nico-famedly: (I mostly had this thought because coming back to this after a week and having to write my own derived tests was a lot harder than necessary because of this looping :))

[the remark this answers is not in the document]
jason-famedly: I think calling them parallel tests is what confused me the most. Should be stripped down now. I didn't see any more but I may have missed them

tests/test_local_joins.py
```diff
@@ -45,61 +48,159 @@ def prepare(self, reactor: MemoryReactor, clock: Clock, homeserver: HomeServer):
...
-            expect_code=HTTPStatus.FORBIDDEN,
-            tok=self.access_token_c,
-        )
-        self.helper.join(
-            room_id,
-            self.user_d,
-            expect_code=HTTPStatus.FORBIDDEN,
-            tok=self.access_token_d,
-        )
+        for user, token in (
```
nico-famedly, line 94: can you remove this looping as well?

nico-famedly (dismissed):
Apart from a few excessive loops, this looks good to me.

tests/test_local_joins.py
jason-famedly: whoops, missed those too

jason-famedly: Should be sorted now

nico-famedly (approved):

PR #77, opened 2025-04-15
fix: with 2 parts

jason-famedly:
1. handle the federation list missing during startup/when accessing permissions(for the permission object event type to use)
2. Better permission validation when existing data is invalid


tests/test_permissions.py
```diff
@@ -696,3 +700,193 @@ def test_invite_before_login(self) -> None:
...
+            self.module_api.account_data_manager.get_global(
+                self.user_c, config_type.value
+            )
+        )
+        assert existing_data is not None, "Data should have been populated"
+        # self._patcher1.start()
+
+
+class PermissionsAccountDataValidationEdgeTestCases(FederatingModuleApiTestCase):
+    """
```
jason-famedly, line 785: I just realized this doesn't test against an EPA mode server. I don't think the code paths should differentiate in a way that matters but if you want that included give me a heads up

nico-famedly: That's fine.

synapse_invite_checker/permissions.py
```diff
@@ -79,11 +100,23 @@ async def get_config_type_from_mxid(
...
+            logger.warning(
+                "Federation list was not available, falling back to local configuration. Reason: %r",
+                e,
+            )
+            if self.config.tim_type == TimType.EPA:
+                config_type = PermissionConfigType.EPA_ACCOUNT_DATA_TYPE
+            else:
+                # Since we default to PRO, this seems prudent unless a better option
+                config_type = PermissionConfigType.PRO_ACCOUNT_DATA_TYPE
+
```
nico-famedly, line 119: I think we should just always use the config value here.

jason-famedly: Yeah, I thought somewhere in the ancient past we talked about not trusting the setting at all and always using the List. I do not recall the specifics of that conversation, but in light of several months of practice, I'm on board

nico-famedly (approved):

nico-famedly (approved):

PR #80, opened 2025-05-03
feat: Add option for https fed list without mtls, reduce clamped room scan min interval

emgrav:
Closes https://github.com/famedly/product-management/issues/3086

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -186,7 +186,7 @@ BASE_API_PREFIX = "/_synapse/client/com.famedly/tim"
 
 
 class InviteChecker:
-    __version__ = "0.4.2"
+    __version__ = "0.4.3"
 
     def __init__(self, config: InviteCheckerConfig, api: ModuleApi):
         self.api = api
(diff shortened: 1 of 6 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -130,15 +130,21 @@ def __init__(self, config: InviteCheckerConfig):
...
             client_cert = PrivateCertificate.loadPEM(content)
-        elif self.url.scheme != "http":
-            msg = "No mtls cert and scheme is not http"
+        elif self.url.scheme == "https" and config.federation_list_require_mtls:
+            # HTTPS with required mTLS but no certificate
+            msg = "No mtls cert and scheme is https with mTLS required"
+            raise Exception(msg)
+        elif self.url.scheme != "http" and self.url.scheme != "https":
+            # Neither HTTP nor HTTPS
+            msg = "URL scheme must be either http or https"
```
nico-famedly, line 147: Can you add a test for this exception? Otherwise jason will yell at me again for not being strict enough :)

nico-famedly (approved):

nico-famedly (dismissed):

nico-famedly (dismissed):

jason-famedly:
The coverage is missing that Exception because the condition it checks can never be reached, it is caught during `parse_config()` before the `InviteChecker` can be initialized. There is nothing that can be done about that, so don't sweat the coverage gap(even tho it is sadness)

nico-famedly (dismissed):

nico-famedly (approved):

PR #82, opened 2025-05-06
fix: Only re-fetch the Federation List if the domain was not found as an entry

jason-famedly:
Not if the domain was found but was not an 'insured' domain

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -15,7 +15,7 @@
 import base64
 import functools
 import logging
-from collections.abc import Callable, Collection
+from collections.abc import Collection
 from contextlib import suppress
 from typing import Any, Literal
 from urllib.parse import quote, urlparse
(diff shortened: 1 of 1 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -499,10 +499,13 @@ async def _fetch_federation_list(
...
-    async def _domain_list_check(self, check: Callable[[str], bool]) -> bool:
-        """Run a `check` against data found on the FederationList"""
+    async def is_domain_allowed(self, domain: str) -> bool:
+        """
+        See if a domain is found on the Federation List. If it was not, refetch the list
+        to try again.
+        """
         fed_list = await self._fetch_federation_list()
-        if check(fed_list):
+        if fed_list.allowed(domain):
```
nico-famedly, line 508: the federation list should be refetched, if the domain is not in the federation list. Also when we are fetching the list to check if the server is an insurance, afaik.

nico-famedly: But the federation list is not refetched, if an insurance is not in the federation list. But I guess it doesn't matter in the cases, that we care about.

nico-famedly (approved):

jason-famedly:
fixes: famedly/product-management#3103

PR #83, opened 2025-05-07
fix: Fix mTLS policy host comparison

emgrav:


synapse_invite_checker/invite_checker.py
```diff
@@ -152,7 +152,16 @@ def __init__(self, config: InviteCheckerConfig):
         )
 
     def creatorForNetloc(self, hostname, port):
+        if isinstance(hostname, bytes):
+            hostname = hostname.decode("utf-8")
```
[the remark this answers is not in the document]
emgrav: (i.e. propagating to the outer scope)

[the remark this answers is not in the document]
nico-famedly: python does not have scopes inside a function. But also I wouldn't do a decode here, since we can't be sure it is utf-8, I think. I would instead encode the url hostname.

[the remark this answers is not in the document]
jason-famedly: Depends on if you nest the function(for my response). In this case it is not, but it will come out of the statement changed

synapse_invite_checker/invite_checker.py
```diff
@@ -152,7 +152,15 @@ def __init__(self, config: InviteCheckerConfig):
         )
 
     def creatorForNetloc(self, hostname, port):
-        if self.url.hostname != hostname or self.url.port != port:
+        if isinstance(hostname, bytes):
```
nico-famedly, line 155: I don't think we need to check, if the input argument is bytes, the docs say it is. the difference was probably the url not being bytes.

jason-famedly: if you think it helps, could probably change the signature line to:
```python
def creatorForNetloc(self, hostname: bytes, port: int) -> IOpenSSLClientConnectionCreator:
```
which is just a nit, and I would say not super critical

nico-famedly (approved):

tests/test_config.py
```diff
@@ -260,3 +260,62 @@ def test_invalid_url_scheme_raises(self) -> None:
 
         config.federation_list_url = "file:///etc/passwd"
         self.assertRaises(Exception, MtlsPolicy, config)
+
+    def test_mtls_policy_creator_for_netloc_string_hostname(self) -> None:
```
nico-famedly, line 264: this testcase makes no sense, since twisted should always pass bytes :)

synapse_invite_checker/invite_checker.py
```diff
@@ -152,7 +152,14 @@ def __init__(self, config: InviteCheckerConfig):
         )
 
     def creatorForNetloc(self, hostname, port):
-        if self.url.hostname != hostname or self.url.port != port:
+        if self.url.hostname.encode("utf-8") != hostname or self.url.port != port:
```
jason-famedly, line 155: Just to make sure, the `self.url.hostname` will *never* come in as bytes from a `ParseResultsBytes` object, right?

jason-famedly: I guess since it gets parsed in from the yaml loader, it will probably always be just a regular string. Fair enough

jason-famedly (dismissed):
Looks good, and thanks for the test!!

jason-famedly (approved):

PR #85, opened 2025-05-12
fix: Insure that local user actually exists prior to retrieving their permissions

jason-famedly:


synapse_invite_checker/invite_checker.py
```diff
@@ -809,6 +809,10 @@ async def user_may_invite(
         # Step 2, check invite settings
         # Skip remote users as we can't check their account data
         if self.api.is_mine(invitee):
+            if not await self.api.check_user_exists(invitee):
+                logger.debug("Local invitee (%s) is not a user that exists", invitee)
```
nico-famedly, line 813:
```suggestion
                logger.debug("Blocking invite to non-existent local user '%s'", invitee)
```

tests/test_remote_invites.py
```diff
@@ -195,6 +196,37 @@ def test_invite_from_remote_outside_of_fed_list(self) -> None:
                     == errors.Codes.FORBIDDEN
                 ), f"'{remote_user_id}' should be FORBIDDEN to invite {local_user}(after permission)"
 
+    def test_invite_from_nonexistent_local_user(self) -> None:
```
nico-famedly, line 199: From? Isn't it "to"?

jason-famedly: whoops, yes good catch

nico-famedly (approved):

PR #86, opened 2025-05-16
fix: Account for invites sent when calculating when to kick local users from an EPA room

jason-famedly:
Fixes: 
* famedly/product-management#3160

Adds a new sub-option in `insured_only_room_scan` to grant a separate grace period to invites

```
insured_only_room_scan:
  invite_grace_period: "1w"
```
Uses the same Duration Parsing and default as `grace_period`

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ modules:
         insured_only_room_scan:
           enabled: true or false  # optional switch to disable the insured-only room scan from running.  The scan is enabled by default, but only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
           grace_period: see 'Duration Parsing' below, # Length of time a room with only EPA members is allowed to exist before deletion. Ignored if `enabled` is false. Defaults to "1w"
+          invite_grace_period: see 'Duration Parsing' below, # A separate grace period just for invites, to allow more time for PRO server users to respond to insured users. Ignored if `enabled` is false. Defaults to "1w"
         inactive_room_scan:
           enabled: true or false # optional switch to disable the room scan for inactive rooms, defaults to true
           grace_period: see 'Duration Parsing' below # Length of time a room is allowed to have no message activity before it is eligible for deletion. Ignored if 'enabled' is false. Defaults to "26w" which is 6 months
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -77,6 +77,7 @@ from synapse_invite_checker.config import DefaultPermissionConfig, InviteChecker
 from synapse_invite_checker.permissions import (
     InviteCheckerPermissionsHandler,
 )
+from synapse_invite_checker.responses import EpaRoomTimestampResults
 from synapse_invite_checker.rest.messenger_info import (
     INFO_API_PREFIX,
     MessengerFindByIkResource,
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -55,6 +55,7 @@ modules:
         insured_only_room_scan:
           enabled: true or false  # optional switch to disable the insured-only room scan from running.  The scan is enabled by default, but only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
           grace_period: see 'Duration Parsing' below, # Length of time a room with only EPA members is allowed to exist before deletion. Ignored if `enabled` is false. Defaults to "1w"
+          invite_grace_period: see 'Duration Parsing' below, # A separate grace period just for invites, to allow more time for PRO server users to respond to insured users. Ignored if `enabled` is false. Defaults to "1w"
```
nico-famedly, line 58: can I set this to infinite?

synapse_invite_checker/invite_checker.py
```diff
@@ -1132,30 +1149,32 @@ async def room_scan(self) -> None:
...
+                        room_id
                     )
+                    current_time = self.api._hs.get_clock().time_msec()
+
+                    if (
+                        epa_room_result.should_kick_because_leave(current_time)
+                        or epa_room_result.should_kick_because_invite(current_time)
+                        or epa_room_result.should_kick_because_no_activity_since_creation(
+                            current_time
+                        )
```
nico-famedly, line 1162: this logic seems backwards, it should be like this:

kick if create event, leave event and invite are older than their respective limits. You can fold create and leave in together, invite should only be considered if there is an invite.

README.md
jason-famedly: :thinking: No. Should you? Infinite isn't really a definable time :laughing: 

That being said, whatever you set here will be bounded by the procedure of the room being purged anyway as an inactive room. Might could do a `0` option like for the room scan disabler? Just so you can turn it off

nico-famedly:
commit message.

synapse_invite_checker/invite_checker.py
```diff
@@ -1132,30 +1151,26 @@ async def room_scan(self) -> None:
...
+                    epa_room_result = await self.define_criteria_to_kick_users_epa_mode(
+                        room_id
                     )
+                    current_time = self.api._hs.get_clock().time_msec()
 
-                    if last_user_left_timestamp is None:
-                        # This is our sign of room either mid-creation/broken
-                        # Just skip this room, as there are no events to change to leave
-                        # and it will be purged automatically below
+                    if epa_room_result.judge_room_for_inactivity(current_time):
```
nico-famedly, line 1159: You aren't checking the room for inactivity. That happens only in the other room scan. Here you check for membership with a grace period.

jason-famedly: `judge_room_for_membership()`? or maybe `judge_room_for_participation()`?

synapse_invite_checker/invite_checker.py
```diff
@@ -1132,30 +1151,26 @@ async def room_scan(self) -> None:
 
                 # only shut down rooms that only have EPA hosts in them
                 if await self.have_all_pro_hosts_left(room_id):
-                    last_user_left_timestamp: int | None = (
-                        await self.get_timestamp_from_eligible_events_for_epa_room_purge(
-                            room_id
-                        )
+                    epa_room_result = await self.define_criteria_to_kick_users_epa_mode(
```
nico-famedly, line 1154: the old name was probably better, apart from that it was timestamp instead of timestamps.

synapse_invite_checker/responses.py
```diff
@@ -0,0 +1,91 @@
...
+            room. Only used if there was no detected 'leave' event.
+        last_leave_in_room: The timestamp of the newest 'leave' membership in the room.
+            The preferred timestamp to acquire.
+        room_creation_ts: If nothing else, use the room creation as a sentinel. The slim
+            possibility exists that no 'leave' event exists and any 'invite' may have
+            failed. This is allowed to give a user time to try again before they do not
+            have access to this room removed.
+    """
+
+    config: InsuredOnlyRoomScanConfig
```
nico-famedly, line 42: if you make this just a basic class and don't put the config in here, you can put it in the types.py. I don't think you need to make this class this fancy, since it is only used in one place.

synapse_invite_checker/invite_checker.py
jason-famedly: guess it doesn't matter, it's gone now

PR #88, opened 2025-06-20
chore: Add mypy config and fix mypy errors

itsoyou:
Ticket: Enable more linting in invite checker
[#2949](https://github.com/famedly/product-management/issues/2949)

It is including a config on `pyproject.toml`

- [ ] Fix the errors reported by `hatch fmt` or add appropriate suppressions with proper comments
- [x] Enable mypy and fix reported type errors: Fix the errors reported by `hatch run types:check` or add appropriate suppressions with proper comments
- [ ] Add both of those linters to our CI jobs

I will continue work on more fixes for `hatch fmt`, and then work on the Github actions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,14 @@ matrix.database.env-vars = [
   { key = "SYNAPSE_POSTGRES_PASSWORD", value = "postgres", if = ["postgres"] },
 ]
 
+[tool.hatch.envs.types]
+extra-dependencies = [
+  "mypy>=1.0.0",
+]
+
+[tool.hatch.envs.types.scripts]
+check = "mypy --ignore-missing-imports --install-types --non-interactive {args:synapse_invite_checker}"
+
 [tool.coverage.run]
 branch = true
 parallel = true
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -17,7 +17,7 @@ import functools
 import logging
 from collections.abc import Collection
 from contextlib import suppress
-from typing import Any, Literal, cast
+from typing import Any, Literal, cast, Mapping
 from urllib.parse import quote, urlparse
 
 from cachetools import TTLCache, keys
@@ -70,7 +70,7 @@ from synapse.util.metrics import measure_func
 from twisted.internet.defer import Deferred
 from twisted.internet.ssl import PrivateCertificate, optionsForClientTLS, platformTrust
 from twisted.web.client import HTTPConnectionPool
-from twisted.web.iweb import IPolicyForHTTPS
+from twisted.web.iweb import IPolicyForHTTPS, IAgent
 from zope.interface import implementer
 
 from synapse_invite_checker.config import DefaultPermissionConfig, InviteCheckerConfig
@@ -153,9 +153,7 @@ class MtlsPolicy:
         )
 
     def creatorForNetloc(self, hostname: bytes, port: int):
-        assert self.url.hostname is not None
-
-        if self.url.hostname.encode("utf-8") != hostname or self.url.port != port:
+        if self.url.hostname and self.url.hostname.encode("utf-8") != hostname or self.url.port != port:
             logger.error(
                 "Destination mismatch: %r:%r != %r:%r",
                 self.url.hostname,
@@ -183,13 +181,16 @@ class FederationAllowListClient(BaseHttpClient):
         super().__init__(hs)
 
         pool = HTTPConnectionPool(self.reactor)
-        self.agent = ProxyAgent(
+
+        mstls_policy = MtlsPolicy(config)
+        proxy_agent = ProxyAgent(
             self.reactor,
             hs.get_reactor(),
             connectTimeout=15,
-            contextFactory=MtlsPolicy(config),
+            contextFactory=cast(IPolicyForHTTPS, mstls_policy),
             pool=pool,
         )
+        self.agent = cast(IAgent, proxy_agent)
 
 
 BASE_API_PREFIX = "/_synapse/client/com.famedly/tim"
@@ -205,9 +206,10 @@ class InviteChecker:
         self.config = config
         # Can not do this as part of parse_config() as there is no access to the server
         # name yet
-        self.config.default_permissions.maybe_update_server_exceptions(
-            self.api._hs.config.server.server_name
-        )
+        if self.config.default_permissions:
+            self.config.default_permissions.maybe_update_server_exceptions(
+                self.api._hs.config.server.server_name
+            )
 
         self.federation_list_client = FederationAllowListClient(api._hs, self.config)
 
@@ -244,12 +246,13 @@ class InviteChecker:
 
         # Make sure this doesn't get initialized until after the default permissions
         # were potentially modified to account for the local server template
-        self.permissions_handler = InviteCheckerPermissionsHandler(
-            self.api,
-            self.config,
-            self.is_domain_insurance,
-            self.config.default_permissions,
-        )
+        if self.config.default_permissions:
+            self.permissions_handler = InviteCheckerPermissionsHandler(
+                self.api,
+                self.config,
+                self.is_domain_insurance,
+                self.config.default_permissions,
+            )
 
         self.task_scheduler = api._hs.get_task_scheduler()
 
@@ -499,7 +502,7 @@ class InviteChecker:
 
         chain = load_certificate(
             FILETYPE_ASN1,
-            await self._raw_gematik_intermediate_cert_fetch(jwskey.get_issuer().CN),
+            await self._raw_gematik_intermediate_cert_fetch(jwskey.get_issuer().CN or ""),
         )
         store_ctx = X509StoreContext(store, jwskey, chain=[chain])
         store_ctx.verify_certificate()
@@ -588,7 +591,7 @@ class InviteChecker:
 
     async def check_event_allowed(
         self, event: EventBase, context: StateMap[EventBase]
-    ) -> tuple[bool, dict | None] | None:
+    ) -> tuple[bool, dict | None]:
         """
         Check the 'event' to see if it is allowed to exist. This takes place before
         the event is actually stored.
@@ -622,9 +625,10 @@ class InviteChecker:
                 return False, None
             creation_event = context.get((EventTypes.Create, ""))
             # `m.federate` defaults to True if unspecified
-            federated_flag = creation_event["content"].get(
-                EventContentFields.FEDERATE, True
-            )
+            if creation_event and creation_event["content"]:
+                federated_flag = creation_event["content"].get(
+                    EventContentFields.FEDERATE, True
+                )
             # Remember to account for the override disabler
             if federated_flag and self.config.override_public_room_federation:
                 return False, None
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -66,13 +67,22 @@ matrix.database.env-vars = [
...
+check = "mypy"
+
 [tool.coverage.run]
 branch = true
 parallel = true
 omit = ["tests/*"]
 
 [tool.ruff]
 target-version = "py311"
+line-length = 88
```
jason-famedly, line 85: Just as a question: what led you to use 88 here and not something larger? Famedly doesn't appear to have any guidance or company policy for this, so I'm not against it

pyproject.toml
```diff
@@ -81,7 +91,10 @@ ignore = [
   "TRY002",
   "TRY003",
   "PLW0603",
-  "N802"
+  "N802",
+  "N815",
+  "PT", # Don't include PT (pytest-style)
+  "RUF013" # allow implicit `Optional`(pressing PEP 484)
```
jason-famedly, line 97: `PT` is rather broad here, is there a narrower option?
And do we still need `RUF013`? I thought getting rid of those `None`s in `PermissionConfig` took care of that?

synapse_invite_checker/invite_checker.py
```diff
@@ -574,21 +586,27 @@ async def user_may_join_room(
...
-            ), "room_data(RoomsForUser) was None after an invite"
-            likely_invite_event_id = room_data.event_id
+            if room_data is None:
+                logger.warning(
+                    "room_data(RoomsForUser) was None after an invite for user (%s) in room (%s)",
+                    user,
+                    room_id,
+                )
+            if room_data:
+                likely_invite_event_id = room_data.event_id
```
jason-famedly, line 596: So the point here was that the `room_data` should never be `None` at this point(hence the huge comment block above). Logically, I'm ok with this change, but now my IDE thinks: `likely_invite_event_id` 'might be referenced before assignment'. Perhaps the rest of the lower block should be indented as well? What do you think?

Maybe could move the `if room_data` part above the `if room_data is None` bit, then `if room_data is None` could just be an `else`?

(As for the codecov complaining about the warning not be covered, I'm not going to ask you to write a test to pry that behavior out. We have the logging for the purpose of catching Synapse being broken somehow, and not our bending over backwards to handle it. Yelling about it in the logs should be enough, I think.)

synapse_invite_checker/invite_checker.py
```diff
@@ -801,20 +818,20 @@ async def user_may_invite(
...
-            state_mapping: StateMap[EventBase] = (
-                await self.api._storage_controllers.state.get_current_state(
-                    room_id,
-                    StateFilter.from_types([(EventTypes.JoinRules, None)]),
-                )
+            state_mapping: StateMap[
+                EventBase
+            ] = await self.api._storage_controllers.state.get_current_state(
+                room_id,
+                StateFilter.from_types([(EventTypes.JoinRules, None)]),
```
jason-famedly, line 825: This is that ruff formatting weirdness, isn't it? I really don't like how that looks, eww. I guess I can live with it if we don't use Black anymore. There may be an issue on the Ruff format tracker thing, I'll see if I can find it since there doesn't seem to be an option to have it like the former.

synapse_invite_checker/permissions.py
```diff
@@ -57,7 +56,7 @@ async def get_permissions(self, user_id: str) -> PermissionConfig:
         account_data = await self.account_data_manager.get_global(
             user_id, config_type.value
         )
-
+        permissions: PermissionConfig | DefaultPermissionConfig
```
jason-famedly, line 59: Just for some information, I know we talked about this a bit out-of-band. (This kind of thing fascinates me :smile: )

I was a bit wrong when I mentioned they may not be related classes, `DefaultPermissionConfig` *is* actually a subclass of `PermissionConfig`. So, technically it should have been perfectly happy accepting this without the extra Type hinting(it's called the "Liskov Substitution Principle"). We'll leave this as is now, but it may be a subtle(and not show stopping) bug in mypy.

pyproject.toml
itsoyou: I think this pr is big enough, so I will create a separate one! :)

synapse_invite_checker/types.py
```diff
@@ -142,29 +142,29 @@ class FederationDomain(BaseModel):
...
 class FederationList(BaseModel):
     model_config = ConfigDict(
         strict=True, frozen=True, extra="ignore", allow_inf_nan=False
     )
 
-    domainList: list[FederationDomain]  # noqa: N815
+    domainList: list[FederationDomain]
 
-    @computed_field
+    @computed_field  # type: ignore[prop-decorator]
```
jason-famedly, line 159: It might be better to try swapping the order of the decorators instead, looks like it just doesn't like that the `property` bit is on the inside. Then you won't have to do a `type: ignore` on them
(Same for the other two, I think)

synapse_invite_checker/invite_checker.py
```diff
@@ -153,9 +156,11 @@ def __init__(self, config: InviteCheckerConfig):
         )
 
     def creatorForNetloc(self, hostname: bytes, port: int):
-        assert self.url.hostname is not None
-
-        if self.url.hostname.encode("utf-8") != hostname or self.url.port != port:
+        if (
+            self.url.hostname
```
jason-famedly, line 160: What do you think the chances are of this being an empty string?

synapse_invite_checker/invite_checker.py
```diff
@@ -205,9 +213,10 @@ def __init__(self, config: InviteCheckerConfig, api: ModuleApi):
         self.config = config
         # Can not do this as part of parse_config() as there is no access to the server
         # name yet
-        self.config.default_permissions.maybe_update_server_exceptions(
-            self.api._hs.config.server.server_name
-        )
+        if self.config.default_permissions:
+            self.config.default_permissions.maybe_update_server_exceptions(
+                self.api._hs.config.server.server_name
+            )
```
jason-famedly, line 219: So this is weird. The `self.config.default_permissions` should never be `None` because it should always exist, logically. I will have to think on it

tests/test_local_invites.py
```diff
@@ -12,6 +12,8 @@
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+# ruff: noqa: ARG002
```
jason-famedly, line 16: Again, globally I'm on the fence for.

As an experiment, let's try pre-pending all the `label` arguments that are unused with a `_`, so make them `_label` instead and see if that makes the thing happy?

synapse_invite_checker/types.py
```diff
@@ -49,14 +49,14 @@ class PermissionConfig(BaseModel):
     # InviteCheckerPermissionsHandler.get_permissions() for where it is set now.
     # TLDR: this will not be an important detail after migrations have run as it will
     # always be set during class instantiation
-    defaultSetting: PermissionDefaultSetting = None
+    defaultSetting: PermissionDefaultSetting = PermissionDefaultSetting.BLOCK_ALL
```
jason-famedly, line 52: Not sure we want to set this to block all. Perhaps just do not set it at all, since it is at the top of the class and it is explicitly set on initiation now(in `get_permissions()`, I believe)?

Come to think of it, need to touch a few things now that `None` is no longer a factor...

pyproject.toml
itsoyou: Black defualt line length is 88, and I wanted ruff to follow it as well.

synapse_invite_checker/types.py
itsoyou: hmm but for me, the error has been gone with the way you suggested. 🧐

synapse_invite_checker/types.py
itsoyou: I set the default as BLOCK_ALL, because I initially thought that common security principal shouldn't follow ALLOW_ALL unless it is explicitly set as like that. Also all the test cases `defaultSetting: allow all` was there, but if we have specific usecases, we can stick to None.

synapse_invite_checker/invite_checker.py
itsoyou: This is not applicable anymore with the patch you've provided. so I am closing it :)

synapse_invite_checker/types.py
jason-famedly: I thought so too. And once upon a time, it was. They changed how the gematik spec worked several months back, and I'm pretty sure that's why it's ALLOW_ALL now.

I've been poking at this, and I think I have some changes that will just move where this is declared in `parse_config()` to be here instead. Still working some kinks in it.

synapse_invite_checker/invite_checker.py
```diff
@@ -653,7 +672,7 @@ async def on_create_room(
         """
         # preset can be any of "private_chat", "trusted_private_chat" or "public_chat"
         # Do not allow "public_chat". Default is based on setting of visibility
-        room_preset: str = request_content.get("preset")
+        room_preset: str = request_content.get("preset", "")
```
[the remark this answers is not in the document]
itsoyou: would this be okay? @jason-famedly 

```suggestion
        # visibility can be either "public" or "private". If not included, it defaults to "private"
        room_visibility: str = request_content.get("visibility", "private")
        # preset can be any of "private_chat", "trusted_private_chat" or "public_chat"
        # Do not allow "public_chat". Default is based on setting of visibility
        preset: str = request_content.get("preset", "")
        # If preset is not specified, infer it from visibility
        if not preset:
            if room_visibility == "public":
                preset = RoomCreationPreset.PUBLIC_CHAT
            else:
                preset = RoomCreationPreset.PRIVATE_CHAT
            request_content["preset"] = preset
        room_preset: str = preset
```

synapse_invite_checker/invite_checker.py
```diff
@@ -574,21 +580,27 @@ async def user_may_join_room(
...
             invite_room_state = invite_event.unsigned.get("invite_room_state", [])
             for _event in invite_room_state:
-                if _event["type"] == EventTypes.JoinRules:
-                    if _event["content"]["join_rule"] == JoinRules.PUBLIC:
-                        return errors.Codes.FORBIDDEN
+                if (
+                    _event["type"] == EventTypes.JoinRules
+                    and _event["content"]["join_rule"] == JoinRules.PUBLIC
+                ):
+                    return errors.Codes.FORBIDDEN
```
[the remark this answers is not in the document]
jason-famedly: That looks great, nice job on the consolidation too!

synapse_invite_checker/invite_checker.py
```diff
@@ -734,10 +751,10 @@ async def on_create_room(
...
         self,
         user_id: str,
-        device_id: str | None,
-        initial_display_name: str | None,
-        request_info: Collection[tuple[str | None, str]],
-        auth_provider_id: str | None = None,
+        device_id: str | None,  # noqa: ARG002
+        initial_display_name: str | None,  # noqa: ARG002
+        request_info: Collection[tuple[str | None, str]],  # noqa: ARG002
+        auth_provider_id: str | None = None,  # noqa: ARG002
```
jason-famedly, line 757: I think we can do the `_` thing on the front of these too, to make the `ARG002`'s go away?

pyproject.toml
```diff
@@ -52,8 +52,8 @@ dependencies = [
 
 [tool.hatch.envs.default.scripts]
 cov = "pytest --cov-report=term-missing --cov-config=pyproject.toml --cov=synapse_invite_checker --cov=tests"
-fmt = "ruff check . --fix"
 format = "black ."
+lint = "ruff check . --fix"
```
jason-famedly, line 56: Good call, forgot about that maybe being confused. I know I sure would. I didn't even know the other `fmt` thing existed until I was told about it  :man_facepalming:

PR #89, opened 2025-06-26
feat: Allow reactions only containing a single emoji

jason-famedly:
Fixes: #famedly/product-management#3239

Still need to:
* [x] write a bit more testing:
   * [x] disabling works
   * [x] no reactions
   * [x] sequences of reactions)
* [x] fix up Synapse to import the module as well, or everything crashes(this imports when installing the SIC, so it's taken care of locally here, not in Synapse)

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2020,2023 Famedly
+# Copyright (C) 2020,2025 Famedly
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
@@ -14,6 +14,8 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 import base64
 import functools
+
+import emoji
 import logging
 from collections.abc import Collection
 from contextlib import suppress
@@ -431,6 +433,13 @@ class InviteChecker:
             "block_invites_into_dms", _config.block_invites_into_dms
         )
         _config.block_invites_into_dms = _block_invites_into_dms
+
+        _limit_reactions = config.get("limit_reactions", _config.limit_reactions)
+        if not isinstance(_limit_reactions, bool):
+            msg = "`limit_reactions` must be a boolean"
+            raise ConfigError(msg)
+        _config.limit_reactions = _limit_reactions
+
         return _config
 
     def after_startup(self) -> None:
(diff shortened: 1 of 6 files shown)


tests/test_reactions.py
```diff
@@ -0,0 +1,164 @@
...
+        message_1_body = self.helper.send(
+            room_id, "message 1", tok=self.map_user_id_to_token[self.user_a]
+        )
+        self.helper.send_event(
+            room_id,
+            EventTypes.Reaction,
+            {
+                "m.relates_to": {
+                    "event_id": message_1_body.get("event_id"),
+                    "key": "H",
```
nico-famedly, line 78: This is not an emoji and as such should be rejected, afaik.

jason-famedly: That's fair. With all the back and forth about it over the last few months I may have gotten confused that *words* and *letters* were not ok but a single emoji was. Nice catch

tests/test_reactions.py
```diff
@@ -0,0 +1,196 @@
...
+        self.helper.send_event(
+            room_id,
+            EventTypes.Reaction,
+            {
+                "m.relates_to": {
+                    "event_id": message_2_body.get("event_id"),
+                    # Two separate emoji are not allowed(in the same reaction)
+                    # :+1: and :smiley:
+                    # \uD83D\uDC4D\uD83D\uDE03
+                    "key": "👍😃",
```
nico-famedly, line 186: can you test for a flag and a family emoji as well to validate, that those are allowed to be sent? (They are composed emojis)

synapse_invite_checker/invite_checker.py
```diff
@@ -610,11 +619,39 @@ async def check_event_allowed(
...
+                # check that it is an emoji(which will exclude normal numbers and
+                # letters) and then check that there is only a single emoji(while
+                # accommodating the aforementioned emoji variations). `emoji_list()`
+                # does seem to handle the variations correctly(based on some limited
+                # testing)
+                if not emoji.purely_emoji(key) or len(emoji.emoji_list(key)) > 1:
+                    # Normally with this API, it is expected to return a bool to
+                    # indicate a failure to comply, however this raises a 403 error with
+                    # the FORBIDDEN code, and gematik wants it to be a 400 with BAD_JSON.
+                    raise SynapseError(
```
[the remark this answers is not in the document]
jason-famedly: Third party callback - [check_event_allowed](https://element-hq.github.io/synapse/latest/modules/third_party_rules_callbacks.html#check_event_allowed)

Spam checker callback - [check_event_for_spam](https://element-hq.github.io/synapse/latest/modules/spam_checker_callbacks.html#check_event_for_spam)

> this directly contradicts the documentation, are you sure about this?

Which part? If you mean about how I said it returns a bool for indication of a failure, that is pretty correct(if not completely 100%). The return typing is `Tuple[bool, Optional[dict]]` and the bool is if it is allowed or not, the optional dict we will not use so it will always be `None`(we will not be changing an event). Or, perhaps you are commenting on my abbreviating the error codes?

If you meant this part:
> If `check_event_allowed` raises an exception, the module is assumed to have failed. The event will not be accepted but is not treated as explicitly rejected, either. An HTTP request causing the module check will likely result in a 500 Internal Server Error.
> ... (Module developers should not use exceptions for rejection.)

It appears that the warning about the 500 error largely stems from what would happen if it was not an explicit `SynapseError`(like if a more generic `ValueError` was raised inside). 
The federation handling side of the equation, which we don't use(when creating the `make_knock` template or when receiving a `send_knock` request) should be safely ignored, as this is neither of those functions and we don't process events not from the local server anyway.
The client side is only in one place, [`create_new_client_event()`](https://github.com/famedly/synapse/blob/5c7d9732a974420eeeb3df1b0afe3fb78253c464/synapse/handlers/message.py#L1311). If I follow what the API said to use(and return `False` in the tuple as required) I will end up with a `403/M_FORBIDDEN` instead of the `400/M_BAD_JSON` that the [gematik](https://gemspec.gematik.de/docs/gemSpec/gemSpec_TI-M_Basis/gemSpec_TI-M_Basis_V1.1.1/#A_26228-01) spec specified. I'm open to suggestion on another way.

> maybe we should use the spam checker API, which has a stricter API contract and allows returning error codes?

The spam checker API version does have the legacy deprecated way to return an error code, and it is customizable to use `M_BAD_JSON` easily enough, but it will be a 403 not a 400. And the main error message would read as `"This message had been rejected as probable spam"`, which doesn't really fit the bill.

nico-famedly (approved):

PR #90, opened 2025-07-02
chore: apply type checks on tests

itsoyou:
## Enable more linting in invite checker [#2949](https://github.com/famedly/product-management/issues/2949)

This is a following pr of the [first one](https://github.com/famedly/synapse-invite-checker/pull/88)

In this PR:

- [x] enable mypy checks in /tests
- [x] fix all mypy errors in /tests
- [x] fixing `ruff check .` errors in /tests
- [x] add isort during formatting

## Important change in this pr:

1. There are no more types environment in hatch.
2. Now we have our own hatch-static-analysis commands to avoid hatch's ruff config, and use the ruff default settings.
3. See readme to learn the command we can use.


pyproject.toml
```diff
@@ -109,6 +119,12 @@ exclude_lines = [
...
 [tool.mypy]
 ignore_missing_imports = true
-install_types = true
-non_interactive = true
-files = ["synapse_invite_checker"]
+namespace_packages = true
+plugins = [
+  "pydantic.mypy",
+  "mypy_zope:plugin"
+]
```
itsoyou, line 126: Thanks to @jason-famedly, who gave me hint about https://github.com/famedly/synapse/blob/master/mypy.ini#L3

tests/server.py
```diff
@@ -495,21 +495,21 @@ def listenUDP(
         return p
 
     def callFromThread(
-        self, callable_: Callable[..., Any], *args: object, **kwargs: object
+        self, callable: Callable[..., Any], *args: object, **kwargs: object
```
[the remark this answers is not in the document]
jason-famedly: Oooooh. I see. That is Twisted's fault for making it so. How ancient of them. Ok nevermind

tests/test_room_message_timestamp.py
```diff
@@ -45,14 +45,14 @@ def test_can_find_last_message_timestamp(self) -> None:
         def send_message_and_assert_latest_activity(room, message, tok) -> None:
             body = self.helper.send(room, message, tok=tok)
 
-            event_id = self.helper.get_event(room, body.get("event_id"), tok=tok)
+            event_id = self.helper.get_event(room, body.get("event_id", ""), tok=tok)
```
jason-famedly, line 48: This one is on me when I was working it. This stuff is misnamed. I blame looking at it for months for everything being blurry :laughing: 
```suggestion
            event_id = body.get("event_id")
            assert event_id is not None
            self.helper.get_event(room, event_id, tok=tok)
```
Something like that makes more sense?

pyproject.toml
```diff
@@ -52,7 +52,10 @@ dependencies = [
 
 [tool.hatch.envs.default.scripts]
 cov = "pytest --cov-report=term-missing --cov-config=pyproject.toml --cov=synapse_invite_checker --cov=tests"
-format = "black ."
+format = [
+  "black .",
+  "ruff check --select I --fix ."
```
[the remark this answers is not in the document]
itsoyou: Actually it is! somehow I kept getting unsorted import error, so I thought it was not default. maybe I just didn't run the lint command at all hehe

pyproject.toml
```diff
@@ -71,11 +74,18 @@ matrix.database.env-vars = [
...
+  "mypy-zope",
+  "types-PyYAML",
+  "types-bleach",
+  "types-cachetools",
+  "types-requests"
 ]
 
 [tool.hatch.envs.types.scripts]
-check = "mypy"
+check = "mypy ."
```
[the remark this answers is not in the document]
itsoyou: It does recursively check all python files. but since we have files = ... configured, we don't need a dot here.

tests/test_utils/fake_room_creation.py
```diff
@@ -77,7 +77,10 @@ def __init__(
         if room_ver is None:
             room_ver = hs_config.server.default_room_version.identifier
 
-        self.room_version = KNOWN_ROOM_VERSIONS.get(room_ver)
+        room_version = KNOWN_ROOM_VERSIONS.get(room_ver)
+        if room_version is None:
```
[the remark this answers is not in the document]
itsoyou: mypy was complaining with `room_version ` could be None, so I had too 🪄

tests/test_scheduled_tasks.py
```diff
@@ -636,15 +695,16 @@ def assert_task_status_for_room_is(
...
-        assert len(completed_task) == (
-            1 if TaskStatus.COMPLETE in status_list else 0
-        ), f"{comment} | completed {completed_task}"
-        assert len(active_task) == (
-            1 if TaskStatus.ACTIVE in status_list else 0
-        ), f"{comment} | active {active_task}"
-        assert len(scheduled_task) == (
-            1 if TaskStatus.SCHEDULED in status_list else 0
-        ), f"{comment} | scheduled {scheduled_task}"
+        if status_list:
```
jason-famedly, line 698: Not completely sure why I made the argument type for `status_list` optionally `None`, all the places I used it had an empty list passed in :thinking: 

Maybe let's just try changing that in the signature? I want to hear your opinion on it before you do it though

pyproject.toml
```diff
@@ -53,7 +53,7 @@ dependencies = [
 [tool.hatch.envs.default.scripts]
 cov = "pytest --cov-report=term-missing --cov-config=pyproject.toml --cov=synapse_invite_checker --cov=tests"
 format = "black ."
-lint = "ruff check . --fix"
+lint = "ruff check --fix ."
```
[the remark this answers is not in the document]
itsoyou: what `hatch fmt` does is
[1] `ruff check --config .../hatch/env/.internal/hatch-static-analysis/.config/.../ruff_defaults.toml --fix .`
[2] `ruff format --config .../hatch/env/.internal/hatch-static-analysis/.config/.../ruff_defaults.toml .`
which is using the default ruff config of hatch environment, that's why we have 2 different results when we run `hatch run lint`.

but I think we came to the conclusion in the 1st pr, that we don't use `hatch fmt` because we don't want ruff formatting, so we split into `hatch run format` and `hatch run lint`.
https://github.com/famedly/synapse-invite-checker/pull/88#discussion_r2166673354

So we can do either options:
1. configure this and use `hatch fmt`
```
[tool.hatch.envs.hatch-static-analysis]
config-path = "none"
```
2. use `hatch run format` and `hatch run lint` instead.

tests/server.py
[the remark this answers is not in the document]
itsoyou: They are `callable` in twisted: https://docs.twisted.org/en/twisted-24.7.0/api/twisted.internet.interfaces.IReactorFromThreads.html#callFromThread
https://docs.twisted.org/en/twisted-24.7.0/api/twisted.internet.interfaces.IReactorInThreads.html#callInThread,
I will add `# noqa: A002`

tests/server.py
```diff
@@ -916,7 +918,7 @@ def write(self, byt: bytes) -> None:
         # TLSMemoryBIOProtocol) get very confused if a read comes back while they are
         # still doing a write. Doing a callLater here breaks the cycle.
         if self.autoflush:
-            self._reactor.callLater(0.0, self.flush)
+            self._reactor.callLater(delay=0.0, callable=self.flush)
```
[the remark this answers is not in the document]
itsoyou: I think it was fixing the issue but I think it was unnecessary change at the end. The revert will be applied in the following commit.

tests/test_scheduled_tasks.py
jason-famedly: Oh, I meant that the places where `None` would have been appropriate I used an empty list instead. Since the list exists but has no items, it passes right on past the `for` loops just below. Just wasn't sure why I did that

pyproject.toml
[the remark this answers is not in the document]
jason-famedly: Oh I'm not saying lets switch to using `hatch fmt`. We decided against for a good reason. I wonder why mine is using the project settings which was exposing more stuff :thinking: 
```console
jason@ploitari:~/dev/famedly/synapse-invite-checker$ hatch fmt
cmd [1] | ruff check --config /home/jason/.local/share/hatch/env/.internal/hatch-static-analysis/.config/Vw25QWAQ/pyproject.toml --fix .
tests/server.py:498:15: A002 Argument `callable` is shadowing a Python builtin
tests/server.py:512:15: A002 Argument `callable` is shadowing a Python builtin
tests/test_utils/fake_room_creation.py:82:30: EM102 Exception must not use an f-string literal, assign to variable first
Found 3 errors.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
```
evidently, it's because I'm in the installed virtual environment, or something. `default` I think it's called. We should really sort out and define how this is supposed to work for developing on this. We both seem to have two vastly different kinds of setup :sweat_smile:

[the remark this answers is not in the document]
itsoyou: the `hatch/.../ruff_defaults.toml` is actually hatch style ruff not the true ruff default. https://github.com/pypa/hatch/issues/1461
`hatch run lint` is not using the hatch's ruff config, so we get default ruff behavior. So if we don't use `hatch fmt` anymore, we don't need additional settings. but if we still want to check with `hatch fmt`, we can use this setting [`[tool.hatch.envs.hatch-static-analysis]` `config-path = "none"`](https://hatch.pypa.io/latest/config/internal/static-analysis/#no-config) so that `hatch fmt` also use the default ruff behavior.

tests/test_utils/fake_room_creation.py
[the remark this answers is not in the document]
jason-famedly: Good enough for me, as long as it yells about it :smiley:

tests/__init__.py
```diff
@@ -20,17 +20,15 @@
...
 
 def get_invite_checker(config: dict):
     def is_mine(user_id: str):
         return user_id.endswith(":example.org")
 
-    hs = Mock(HomeServer, hostname="example.org")
-
     api = Mock(ModuleApi)
-    api._hs = hs
+    api.hs = Mock(HomeServer, hostname="example.org")
```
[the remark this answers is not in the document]
itsoyou: hmm I don't see any usage of `get_invite_checker` anywhere, same for `admins`.

tests/test_scheduled_tasks.py
```diff
@@ -12,8 +12,11 @@
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
+# mypy: disable-error-code=call-arg
```
jason-famedly, line 15: Oh, that's new. What is that doing?

jason-famedly: Oh. I see what it is. That is probably because of that obnoxious `@cached` decorator on those database calls. It's helpful for quickly pulling repeatedly-the-same data, but makes type checking a pain. That was the purpose of that mypy plugin they use inside of synapse itself, it teaches mypy what that is and how to go around that. We are not doing that, so what you have done is fine. (I would prefer 'by line' for this, but if there are more than 8 then I'm not going to care that much. That we have to do it at all is ridiculous)

pyproject.toml
```diff
@@ -52,10 +52,27 @@ dependencies = [
   "matrix-synapse @ git+https://github.com/famedly/synapse.git@master"
 ]
 
+[tool.hatch.envs.hatch-static-analysis]
+dependencies = ["black", "ruff", "isort"]
+
+[tool.hatch.envs.hatch-static-analysis.scripts]
+format-check = [
+  "black . --check --diff",
+  "isort . --check --diff"
```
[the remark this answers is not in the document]
jason-famedly: Also, um
```console
Usage: black [OPTIONS] SRC ...
```
perhaps the `.` should be on the end?

.github/workflows/python.yml
```diff
@@ -49,7 +49,7 @@ jobs:
             postgres:15
 
       - name: Type checking
-        run: hatch run types:check
+        run: hatch fmt --check -l
```
jason-famedly, line 52: Should probably see if we need any of the ones below this still?

tests/unittest.py
```diff
@@ -354,7 +347,7 @@ def setUp(self) -> None:
             hs=self.hs,
         )
 
-        self.helper = RestHelper(
+        self.helper = RestHelper(  # type: ignore
```
jason-famedly, line 350: So there are a bunch of generic `type: ignore` now, why were these not more specific?

tests/__init__.py
```diff
@@ -12,26 +12,6 @@
...
-
-    hs = Mock(HomeServer, hostname="example.org")
-
-    api = Mock(ModuleApi)
-    api._hs = hs
-    api.server_name = "example.org"
-    api.is_mine.side_effect = is_mine
-
-    return InviteChecker(InviteChecker.parse_config(config), api)
+from synapse_invite_checker import InviteChecker  # noqa: F401
```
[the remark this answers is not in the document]
nico-famedly: You probably want to disable F401 for the whole file?

tests/server.py
```diff
@@ -495,22 +490,34 @@ def listenUDP(
...
     def callFromThread(
-        self, callable_: Callable[..., Any], *args: object, **kwargs: object
+        self,
+        callable: Callable[..., Any],  # noqa: A002
+        *args: object,
+        **kwargs: object,
     ) -> None:
         """
         Make the callback fire in the next reactor iteration.
+        We ignore A002 Argument `callable` is shadowing a Python builtin, because this inherits from
```
nico-famedly, line 500: I would suggest putting that in the A002 comment, but I guess you decided not to, because it was too long?

itsoyou: I could make it shorter and fit into the `# noqa: A002` line

nico-famedly: Alternatively move most of the comment to the line above the argument, I guess?

.github/workflows/python.yml
itsoyou: I've merged type-checking and black into this 
```
      - name: Type checking and Formatting
        run: hatch fmt --check
```
but the linting with ruff remains because it runs different target versions with workflow matrix

tests/utils.py
```diff
@@ -272,7 +272,7 @@ def looping_call(
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> None:
-        self.loopers.append(Looper(function, interval / 1000.0, self.now, args, kwargs))
+        self.loopers.append(Looper(function, interval / 1000.0, self.now, args, kwargs))  # type: ignore[call-arg]
```
nico-famedly, line 275: These call-arg are still needed?

nico-famedly (dismissed):
Looks good to me in general. Minor nitpick left, but we can also merge it as is imo.

.github/workflows/python.yml
jason-famedly: Yes, I suppose there isn't much to be done about that. At least `ruff check` is relatively fast

pyproject.toml
```diff
@@ -71,21 +103,13 @@ matrix.database.env-vars = [
...
-check = "mypy"
-
 [tool.coverage.run]
 branch = true
 parallel = true
 omit = ["tests/*"]
 
 [tool.ruff]
-target-version = "py312"
+target-version = "py313"
```
jason-famedly, line 112: This feels weird. Shouldn't this be the minimum python version that the syntax used supports? I would have thought that would be 3.10. I know for the CI this is overridden by that one action we use. Hopefully this doesn't come back to bite us

jason-famedly (dismissed):
Looks good to me! :rocket:

PR #96, opened 2025-07-29
chore: Rework to not have `ResourceWarnings` during test runs

jason-famedly:
And delete the now unused `InviteCheckerStore`.

This was caused by the "saved" and held open sqlite connection used between tests. In memory sqlite tests are already really fast and our tests are not ridiculously long anyway. Silence these warnings by fixing the underlying cause

<del>This did not affect production use. It was *only* during testing</del>Technically, this is untrue. As part of the initialization of the Invite Checker, a database connection was stolen from the pool early and saved in the(now deleted) `InviteCheckerStore`. While it was used for one of the database calls we now make, that was unnecessary for a call that should only be made on occasion(during the room scan while getting the list of all `room_id`'s). With this released, it not only helps clear some of the `ResourceWarning` spam but also returns that connection to the database pool.

While I am pleased they do not show up in CI, locally the console spam shows as 1000 cumulative lines of:
```
tests/test_createrooms_remote.py::RemoteProModeCreateRoomTest::test_any_user_to_non_fed_domain_create_room_fails_1_private
  /home/jason/dev/famedly/synapse-invite-checker/tests/unittest.py:214: ResourceWarning: unclosed database in <sqlite3.Connection object at 0x72e94e563e20>
    gc.collect(0)

tests/test_createrooms_remote.py::RemoteProModeCreateRoomTest::test_create_room_0_public
  /home/jason/dev/famedly/synapse-invite-checker/tests/unittest.py:214: ResourceWarning: unclosed database in <sqlite3.Connection object at 0x72e94dd7b970>
    gc.collect(0)

tests/test_createrooms_remote.py::RemoteProModeCreateRoomTest::test_create_room_1_private
  /home/jason/dev/famedly/synapse-invite-checker/tests/unittest.py:214: ResourceWarning: unclosed database in <sqlite3.Connection object at 0x72e94d4e4310>
    gc.collect(0)
```
and
```
/home/jason/.local/share/hatch/env/virtual/synapse-invite-checker/Vw25QWAQ/synapse-invite-checker/lib/python3.13/site-packages/_pytest/unraisableexception.py:33: ResourceWarning: unclosed database in <sqlite3.Connection object at 0x72e94c21fa60>
  gc.collect()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
And is frankly unnecessary. Hard to find out what the error is, *if* there is one, through all the noise. Make it go away


itsoyou (approved):
I am super happy that you got it resolved 🎉

PR #99, opened 2025-08-13
fix: Changes required to be used with Synapse `1.136.0`

jason-famedly:
A major rework of the metrics bits were added in for Synapse `1.136.0`. These are bits that allow the invite checker to continue to function

Requires: #98 

Will require the Synapse `v1.136.0_1` release to be on the `master` branch before testing will work correctly


jason-famedly:
I feel like there should be a way to clarify in `pyproject.toml` that this version requires a minimum Synapse version....

nico-famedly (approved):

PR #120, opened 2026-03-17
chore: Add debug logging for unexcepted create room exceptions

nico-famedly:


jason-famedly:
This was superseded by #125 that incorporated more logging than just this

PR #125, opened 2026-03-26
chore: add debug logging for unexcepted callback exceptions

FrenchGithubUser:
SYN-24


jason-famedly (dismissed):
if that blanket `403`/`M_FORBIDDEN` ends up being a problem, we'll just have to fix it up later

jason-famedly (approved):
🚀

PR #128, opened 2026-03-31
feat: disable room shutdowns by default for TIM 1.2 and above

FrenchGithubUser:
SYN-18

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ modules:
           - "10"
         room_scan_run_interval: see 'Duration Parsing' below, # How often to scan for rooms that are eligible for deletion. Defaults to "1h". Setting to "0" completely disables all room scanning
         insured_only_room_scan:
-          enabled: true or false  # optional switch to disable the insured-only room scan from running.  The scan is enabled by default, but only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
+          enabled: true or false  # optional switch to disable the insured-only room scan from running. The scan is enabled by default, but only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
           grace_period: see 'Duration Parsing' below, # Length of time a room with only EPA members is allowed to exist before deletion. Ignored if `enabled` is false. Defaults to "1w"
           invites_grace_period: see 'Duration Parsing' below, # Optional, a separate grace period just for invites, after which an invite will be considered stale and ignored. Otherwise invited "Pro" users are considered joined and will prevent purging the room. Ignored if `enabled` is false. Defaults to "0", which will never consider an invite stale.
         inactive_room_scan:  # This section only applies to TIM version 1.1. If using a different TIM version, these settings will be ignored
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -408,10 +408,19 @@ class InviteChecker:
             msg = "`insured_only_room_scan` should be configured as a dictionary"
             raise ConfigError(msg)
 
-        # Only default enable this room scan if in EPA mode
+        # Defaults to true if in ePA mode and TIM version is 1.1
+        # Defaults to false if TIM version is 1.2 or higher
         enable_insured_room_scan = insured_room_scan_section.get(
-            "enabled", _config.tim_type == TimType.EPA
+            "enabled",
+            (
+                _config.tim_type == TimType.EPA
+                if _config.tim_version < TimVersion.V1_2
+                else False
+            ),
         )
+        if not isinstance(enable_insured_room_scan, bool):
+            msg = "`insured_only_room_scan.enabled` must be a boolean"
+            raise ConfigError(msg)
         # But also prevent it running in PRO mode completely
         enable_insured_room_scan = (
             False if _config.tim_type == TimType.PRO else enable_insured_room_scan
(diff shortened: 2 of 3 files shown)


jason-famedly (commented):
I think it will end up being that we don't get to force this to disabled but still allow for it to function if enabled. That needs to be documented explicitly

README.md
```diff
@@ -55,12 +55,12 @@ modules:
           - "9"
           - "10"
         room_scan_run_interval: see 'Duration Parsing' below, # How often to scan for rooms that are eligible for deletion. Defaults to "1h". Setting to "0" completely disables all room scanning
-        insured_only_room_scan:
-          enabled: true or false  # optional switch to disable the insured-only room scan from running.  The scan is enabled by default, but only runs in EPA mode, otherwise this option is ignored and the scan is disabled.
+        insured_only_room_scan:  # This section only applies to TIM version 1.1. Room shutdowns are prohibited in TIM 1.2 or newer, so this setting is ignored and the scan is disabled for those versions.
+          enabled: true or false  # optional switch to disable the insured-only room scan from running. The scan is enabled by default, but only runs in EPA mode and only for TIM version 1.1, otherwise this option is ignored and the scan is disabled.
```
jason-famedly: Not sure this is right. Insured scans were not mentioned in the gematik spec, so in theory are not supported any more. That doesn't really mean they were prohibited though, I think.

synapse_invite_checker/invite_checker.py
```diff
@@ -371,10 +371,19 @@ def parse_config(config: dict[str, Any]) -> InviteCheckerConfig:
...
-        # Only default enable this room scan if in EPA mode
+        # Defaults to true if in ePA mode and TIM version is 1.1
+        # Defaults to false if TIM version is 1.2 or higher
         enable_insured_room_scan = insured_room_scan_section.get(
-            "enabled", _config.tim_type == TimType.EPA
+            "enabled",
+            (
+                _config.tim_type == TimType.EPA
+                if _config.tim_version < TimVersion.V1_2
+                else False
```
jason-famedly, line 418: Hmm, now we don't get the logging. I'd like to have had that, but I guess it'll be fine.

README.md
```diff
@@ -55,12 +55,12 @@ modules:
...
           grace_period: see 'Duration Parsing' below, # Length of time a room with only EPA members is allowed to exist before deletion. Ignored if `enabled` is false. Defaults to "1w"
           invites_grace_period: see 'Duration Parsing' below, # Optional, a separate grace period just for invites, after which an invite will be considered stale and ignored. Otherwise invited "Pro" users are considered joined and will prevent purging the room. Ignored if `enabled` is false. Defaults to "0", which will never consider an invite stale.
-        inactive_room_scan:  # This section only applies to TIM version 1.1. If using a different TIM version, these settings will be ignored
-          enabled: true or false # optional switch to disable the room scan for inactive rooms, defaults to true if TIM version is set to "1.1"
+        inactive_room_scan:
+          enabled: true or false # optional switch to disable the room scan for inactive rooms, defaults to false if TIM version is "1.2" or higher, defaults to true if TIM version is set to "1.1" and TIM type is ePA
```
jason-famedly: Err, why are we changing this? This has nothing to do with EPA or not unlike the insured rooms

synapse_invite_checker/invite_checker.py
FrenchGithubUser: yea there was no logging before, but it's nice to have one indeed

README.md
FrenchGithubUser: hmm I thought I removed it, might have missclicked something there

tests/test_config.py
```diff
@@ -259,6 +259,12 @@ def test_dict_unexpectedly_is_something_else_raises(self) -> None:
         with pytest.raises(ConfigError):
             InviteChecker.parse_config(test_config)
 
+    def test_insured_only_room_scan_disabled_by_default_for_tim_v1_2(self) -> None:
+        test_config = self.config.copy()
+        test_config.update({"tim-type": "epa", "tim_version": "1.2"})
+        config = InviteChecker.parse_config(test_config)
+        assert not config.insured_room_scan_options.enabled
+
```
jason-famedly: I'm thinking we are leaving some bits on the table instead of working them in here...

How about this: there are 3 tests above that test for setting the tim version. Let's re-work those to test all the defaults we believe are being set by just changing the tim version. Then you can borrow that test to check this value too, and all the assumptions being made about the other parts of the room scan. Does that make sense?

README.md
FrenchGithubUser: I mixed things up in my mind indeed

tests/test_config.py
FrenchGithubUser: yes good idea

tests/test_config.py
```diff
@@ -58,22 +58,46 @@ def test_incorrect_tim_type_raises(self) -> None:
         with pytest.raises(ConfigError):
             InviteChecker.parse_config(test_config)
 
-    def test_tim_version_defaults_to_1_1(self) -> None:
+    def test_tim_v1_1_room_scan_defaults(self) -> None:
+        """
+        When no TIM version is set, it defaults to 1.1. Verify that the room scan
+        options reflect 1.1 defaults: inactive scan enabled, state-only purge and
+        insured scan disabled.
+        """
```
jason-famedly: Let's grab one for tim 1.1 and epa also please

FrenchGithubUser: done

jason-famedly (approved):
🚀

PR #129, opened 2026-04-01
chore: test more room versions in tests, from "9" to "12"

jason-famedly:
SYN-31

Refactor a bunch of tests to be parameterized and check previous and newer room versions. This does have the "less-than-nice" effect of multiplying the number of tests ran

Because the `FakeRoom` does not support room version "11" and "12" yet, this does not touch:
* `OutgoingEPARemoteJoinTestCase`
* `OutgoingPRORemoteJoinTestCase`
* `InactiveRoomScanTaskV1_1TestCase`
* `StateOnlyPurgeRoomScanTaskV1_2TestCase`

These an be handled in a follow-up work, as that is more invasive and this was large enough to look at.

May be best reviewed commit-by-commit

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -17,6 +17,7 @@ import json
 import logging
 from collections.abc import Iterable
 from http import HTTPStatus
+from random import random
 from typing import TYPE_CHECKING, Any, ClassVar
 from unittest.mock import AsyncMock, Mock, patch
 
@@ -216,7 +217,8 @@ class FakeInviteResponse:
 class FederatingModuleApiTestCase(synapsetest.FederatingHomeserverTestCase):
     server_name_for_this_server = SERVER_NAME_FROM_LIST
     OTHER_SERVER_NAME = INSURANCE_DOMAIN_IN_LIST
-    ROOM_VERSION = "10"
+    DEFAULT_ROOM_VERSION = "10"
+    ALLOWED_ROOM_VERSIONS = ["9", "10"]
     TIM_VERSION = TimVersion.V1_1
 
     @classmethod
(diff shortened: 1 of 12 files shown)


jason-famedly:
Just over 30 minutes to complete each test round. That is stretching what I consider acceptable into the "no" category. There is a possibility of refactor into splitting our test runs for each python version into two: One that uses SQLite and one that uses Postgres(see the hatch matrix used in pyproject.toml). This will have to be a separate work though

tests/base.py
```diff
@@ -415,6 +425,16 @@ def send_leave(self, leaving_user: str, room_id: str) -> None:
...
     ) -> str:
+        # Creating a room before version "12" means the room gets a randomly generated
+        # room id. Starting with "12" the room is based on the create event itself. To
+        # avoid arbitrary errors when doing tests that repeatedly create rooms
+        # rapid-fire, let's introduce a small time advance so that the timestamp on the
+        # room is just different enough to have a different hash than a previously
+        # created room. This should help stop errors like
+        #  "store_room with room_id=!dE_jLFdFgPIHXZuc1wZ3qVANUu4R0y8zimGAzqmbqOg failed:
+        #   UNIQUE constraint failed: rooms.room_id".
+        self.reactor.advance(random() * 2)
```
FrenchGithubUser, line 436: why do you use a random value here, and how big can it get?

FrenchGithubUser: and same below

jason-famedly: Each TestCase class gets it's own database copy. But each test function restarts the time to `0`. Since time is screwy and not real in tests, and since the room version "12" uses the hash of the event to calculate what the room ID is, a bunch of rooms were being created with the same hash :laughing: See the above comments. I chose a random value as the timestamp is based on milliseconds so with a window of 2 seconds multiplied by the random, we get a 2000 potential timestamps to avoid the clash of the hashes. Does that make sense?

jason-famedly: If you mean for what `random()` spits out, it's a float value between 0 and 1. Doesn't really matter how many digits it gives, but it's not a ton. Maybe 10 or so?

FrenchGithubUser: okay yea sounds fine :)

FrenchGithubUser (approved):
nice, hopefully this will catch some issues in the future!

PR #131, opened 2026-04-03
chore: Rework testing to reduce time spent and to remove the `hatch run cov` command

jason-famedly:
Tests were taking to long. Let's speed them up!

Before, we ran unit tests and collected coverage with the command `hatch run cov`. This ran the tests on a single process/thread and (in CI) tested against both SQLite and Postgres as a matrix, consecutively.

Now, the built-in `hatch test` command can be used for the same purpose.
`hatch test` will run all tests without coverage on a single process/thread
Add `-p` as an option to run all tests using all available cores/threads of a CPU.
Add `-c` as an option to collect coverage data.

These are interchangeable. Previously, on my machine these tests with coverage took just over 2 minutes, now it is just over 30 seconds. Without coverage is even faster.

SQLite testing will be the default. Running in Postgres is only slightly more labor intensive. A Postgres installation is required of course(see the `python.yml` testing workflow for a quick example) and then add to the front of your command the ENV variables that the testing suite will automatically look for:
`SYNAPSE_POSTGRES=1 SYNAPSE_POSTGRES_USER=<theDatabaseUser> SYNAPSE_POSTGRES_PASSWORD=<theDBUsersPassword> hatch test`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 name = "synapse-invite-checker"
 description = 'Synapse module to handle TIM contact management and invite permissions'
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.10"
 license = "AGPL-3.0-only"
 keywords = []
 authors = [
@@ -15,8 +15,10 @@ authors = [
 classifiers = [
   "Development Status :: 4 - Beta",
   "Programming Language :: Python",
-  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 name = "synapse-invite-checker"
 description = 'Synapse module to handle TIM contact management and invite permissions'
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.10"
```
jason-famedly, line 9: Our test matrix for Python 3.10 was not behaving as expected. The version of Python that was *actually* being used was 3.14. The Github runner has 3.12 installed by default, so this 3.14 was likely being installed by `hatch`.

The theory was that because the version of Python we were trying to use was below this required minimum, `hatch` was *helpfully* installing the newest Python version for us. There was not a warning or an error message, only by happening to notice that the environment that was being produced contained the wrong version string was this caught.

pyproject.toml
```diff
@@ -143,3 +125,57 @@ files = ["synapse_invite_checker", "tests"]
...
+run = "pytest{env:HATCH_TEST_ARGS:} {args}"
+
+# The original 'run-cov' is below, this is overriden to produce coverage properly.
+#run-cov = "coverage run -m pytest{env:HATCH_TEST_ARGS:} {args}"
+# Notes for this command line: Using --cov=synapse_invite_checker instead of having this
+#  directory in the 'coverage.run' config above actually produces the correct coverage
+#  files(the ones that were missing, as mentioned above). The 'cov-report' options both
+#  produce useful output(one for the codecov file and the other displays the missing
+#  coverage lines).
+run-cov = "pytest{env:HATCH_TEST_ARGS:} --cov=synapse_invite_checker --cov-report=lcov:lcov.info --cov-report=term-missing --cov-config=pyproject.toml {args}"
```
jason-famedly, line 166: Of note here(beyond the comments in the file) is that the option to produce an XML file and the coverage of tests has been omitted. We never covered the tests themselves(they were omitted from the coverage results in the above `coverage.run` config), and the XML file was never used(LCOV is used by Codecov instead)

jason-famedly:
This will need the required test checks adjusted before it can be merged, as they will have changed

pyproject.toml
```diff
@@ -83,27 +85,16 @@ lint-fix = [
...
-  { key = "SYNAPSE_POSTGRES_USER", value = "postgres", if = ["postgres"] },
-  { key = "SYNAPSE_POSTGRES_PASSWORD", value = "postgres", if = ["postgres"] },
-]
-
 [tool.coverage.run]
 branch = true
 parallel = true
-omit = ["tests/*"]
+
+[tool.coverage.report]
```
FrenchGithubUser, line 92: why does this need to be moved around?

jason-famedly: Functionally? No reason. I got tired of scrolling around to find the different section while running the experiments and so no reason why the section for the same tool needed to be separated. Traditionally, they are kept together. The toml parser doesn't care, but my eyes did!

FrenchGithubUser (approved):
Looks good to me!

PR #133, opened 2026-05-26
chore: update GitHub Action pins

nikzen:
## Summary
- Update GitHub Actions in workflows to the latest upstream default-branch commit SHAs.
- Keep existing action paths and only replace refs after `@`.

## Verification
- Checked generated diff to include only `.github/**/*.yml` / `.github/**/*.yaml` workflow action refs.
- CI on this PR should validate the updated action versions.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


jason-famedly (approved):
This looks fine.

We need to try and swap out the chartboost/ruff-action for the new built in tooling that does the same thing(this action has not been updated since 2024 and is archived). The migration to a newer codecov action allows using the `oidc` login method, so that should be adjusted as well

PR #137, opened 2026-08-11
chore!: Prohibit history visibility of `world_readable` if a room is allowed to federate

jason-famedly:
SYN-98

BREAKING CHANGE: The configuration option
`prohibit_world_readable_rooms` now defaults to `False`. Behavior has
been adjusted to closely match the original intent of why the default
was `True`.

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -908,33 +908,32 @@ class InviteChecker:
                 )
             # PRO: only being "public" on PRO if "m.federate" is set to True in the creation event
             elif self.config.tim_type == TimType.PRO and join_rule == JoinRules.PUBLIC:
-                creation_event = context.get((EventTypes.Create, ""))
-                # `m.federate` defaults to True if unspecified
-                federated_flag = True
-                if creation_event and creation_event["content"]:
-                    federated_flag = creation_event["content"].get(
-                        EventContentFields.FEDERATE, True
-                    )
                 # Remember to account for the override disabler
-                if federated_flag and self.config.override_public_room_federation:
+                if (
+                    is_room_federate_capable(context)
+                    and self.config.override_public_room_federation
+                ):
                     raise SynapseError(
                         400,
                         "Room cannot be federated",
                         "M_INVALID_ROOM_STATE",
                     )
 
-        # Forbid "m.room.history_visibility" from being "world_readable" when configured
-        # via `prohibit_world_readable_rooms`, or always for EPA servers on TIM >= 1.2.
+        # Forbid "m.room.history_visibility" of "world_readable" when the room is capable of federation,
+        # configured via `prohibit_world_readable_rooms`, or always for EPA servers on TIM >= 1.2.
         elif (
-            (
-                self.config.prohibit_world_readable_rooms
+            event.type == EventTypes.RoomHistoryVisibility
+            and event.content["history_visibility"] == HistoryVisibility.WORLD_READABLE
+            and (
+                (
+                    self.config.prohibit_world_readable_rooms
+                    or is_room_federate_capable(context)
+                )
                 or (
                     self.config.tim_type == TimType.EPA
                     and self.config.tim_version >= TimVersion.V1_2
                 )
             )
-            and event.type == EventTypes.RoomHistoryVisibility
-            and event.content["history_visibility"] == HistoryVisibility.WORLD_READABLE
         ):
             raise SynapseError(
                 400,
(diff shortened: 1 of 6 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -1002,14 +1002,14 @@ async def on_create_room(
...
 
-        # Check if history_visibility is already set in initial_state
-        has_history_visibility = any(
-            event.get("type") == EventTypes.RoomHistoryVisibility
-            for event in initial_state
-        )
+        # Save the data of the history visibility event, if it is found
+        history_visibility_event: dict[str, Any] | None = None
+        for initial_state_event in initial_state:
+            if initial_state_event["type"] == EventTypes.RoomHistoryVisibility:
```
FrenchGithubUser: why did it go from `.get("type")` to `["type"]`?

synapse_invite_checker/invite_checker.py
```diff
@@ -1019,6 +1019,22 @@ async def on_create_room(
...
+            # dangling room. The other is for the instance of an attempt of changing this setting after the room is
+            # already created.
+            if history_visibility_event["content"].get(
+                "history_visibility"
+            ) == HistoryVisibility.WORLD_READABLE and (
+                can_federate or self.config.prohibit_world_readable_rooms
+            ):
+                raise SynapseError(
+                    400,
+                    "History visibility 'world_readable' is not allowed in a room allowed to federate",
```
FrenchGithubUser:
```suggestion
                    "History visibility 'world_readable' is not allowed in a room allowed to federate, or world-readable rooms are forbidden on this server",
```

synapse_invite_checker/invite_checker.py
```diff
@@ -1019,6 +1019,22 @@ async def on_create_room(
...
 
+        else:
+            # A provided history visibility event must not be set to world_readable if the ability to federate is
+            # enabled. This guard has a comparable twin in `check_event_allowed()`, but this one forbids creation of a
+            # dangling room. The other is for the instance of an attempt of changing this setting after the room is
+            # already created.
+            if history_visibility_event["content"].get(
+                "history_visibility"
+            ) == HistoryVisibility.WORLD_READABLE and (
+                can_federate or self.config.prohibit_world_readable_rooms
```
FrenchGithubUser: the value of `can_federate` used here doesn't necessarily reflect what's in the final payload. line 985 might override `m.federate` to `False` after `can_federate` has been computed.

I would say it's fine to enforce a properly formed payload instead of silently overriding it, but wanted to have it written down

jason-famedly: Since this hook runs before the room is actually formed, I wanted to make sure it throws them an error based on the input they gave and not just massage the system to produce them a room that did not meet what they thought they were getting. 

Yes it is overridden above and not updated to verify the original request meets parameters. Largely, that is because `m.federate` being `True` is a default they can forget, whereas `m.room.history_visibility` being `world_readable` is something they actively have to touch. If it was also a default, I would not mind persuading the system to make something close to what they wanted :D 

Hopefully that made sense?

synapse_invite_checker/invite_checker.py
jason-famedly: Slightly cleaner look, otherwise no real reason. 

Should be no procedural difference. I suppose you could argue that it could be `None` and thus throw a `KeyError`. Seems to me that should have been caught by the create room code before it gets here, but maybe it isn't. If it isn't, I guess they get to know about it a little early :D

synapse_invite_checker/invite_checker.py
jason-famedly: Amended relevant commit to include this

synapse_invite_checker/invite_checker.py
jason-famedly: Having given this more than 5 minute of thought, you may have a point here. At least if it is `None` the error is not produced in our code but in the appropriate error handling place that inspects such things. I'll put `.get()` back. Good call

synapse_invite_checker/invite_checker.py
```diff
@@ -1019,6 +1019,23 @@ async def on_create_room(
...
+            ):
```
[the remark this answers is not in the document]
jason-famedly: Good shout

synapse_invite_checker/invite_checker.py
FrenchGithubUser: Yes this is fine, I agree!

synapse_invite_checker/invite_checker.py
jason-famedly: There is another case just like this one below, I fixed that one too. In the last force push/amend of the relevant commit

FrenchGithubUser (approved):
looks sane to me!

PR #138, opened 2026-08-11
fix: Unpin ruff and fix all its errors

jason-famedly:
Depends on #137 
Follow up to the pin of `ruff` in #137 due to version increment causing a lot of warnings/errors

### DIFF
--- a/synapse_invite_checker/invite_checker.py
+++ b/synapse_invite_checker/invite_checker.py
@@ -96,7 +96,6 @@ if TYPE_CHECKING:
     from synapse.storage.roommember import RoomsForUser
 
 # We need to access the private API in some places, in particular the store and the homeserver
-# ruff: noqa: SLF001
 
 
 def cached(cache):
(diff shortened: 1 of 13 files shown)


synapse_invite_checker/invite_checker.py
```diff
@@ -1374,9 +1373,9 @@ async def _get_filtered_events_from_pagination(
         )
 
         (
-            events,
-            next_key,
-            _,
+            list_of_events,
+            _next_key,
+            _limited,
```
jason-famedly, line 1378: We only actually use the one, but this way they are at least labeled

tests/base.py
```diff
@@ -219,7 +218,7 @@ class FederatingModuleApiTestCase(synapsetest.FederatingHomeserverTestCase):
     server_name_for_this_server = SERVER_NAME_FROM_LIST
     OTHER_SERVER_NAME = INSURANCE_DOMAIN_IN_LIST
     DEFAULT_ROOM_VERSION = "10"
-    ALLOWED_ROOM_VERSIONS = ["9", "10"]
+    ALLOWED_ROOM_VERSIONS: ClassVar = ["9", "10", "11", "12"]
```
jason-famedly, line 221: This change is probably contentious. I'm double checking this is tested properly since it is technically against spec, but allows us to forward look at newer room versions. The removals below basically all room tests anyway. The default room version is still pinned to "10", even though there are several tests that also check newer room versions.

Yes, it looks like this is exercised in `test_room_versions.RoomVersionCreateRoomTest`

FrenchGithubUser (approved):
looks fine to me, nice little improvements!

